PR #1, opened 2019-10-28
Animated GIF overlay filter

izzytwosheds:
Implementation of new filter, which overlays animated GIFs on top of a video.
 - Static bitmap and animated GIF filters share a lot of common logic, which is now abstracted into base class. Difference between classes is now how they load a texture from file and which texture ID is rendered onto video frame at a given presentation time. 
 - Using Glide to load animated GIF
 - Animated GIF frames are loaded into a circular buffer, along with their delays (duration of each frame)

### DIFF
--- a/litr-filters/src/main/java/com/linkedin/android/litr/filter/video/gl/AnimatedGifOverlayFilter.java
+++ b/litr-filters/src/main/java/com/linkedin/android/litr/filter/video/gl/AnimatedGifOverlayFilter.java
@@ -7,27 +7,12 @@
  */
 package com.linkedin.android.litr.filter.video.gl;
 
-import android.content.ContentResolver;
-import android.content.Context;
-import android.content.res.AssetFileDescriptor;
 import android.graphics.Bitmap;
 import android.graphics.RectF;
-import android.net.Uri;
 import android.util.Log;
 import androidx.annotation.IntRange;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import com.bumptech.glide.gifdecoder.GifDecoder;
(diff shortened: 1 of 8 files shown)


litr-filters/src/main/java/com/linkedin/android/litr/filter/video/gl/AnimatedGifOverlayFilter.java
```diff
@@ -36,64 +21,55 @@
...
-                if (prevFrame != null) {
-                    prevFrame.next = frame;
-                }
-                prevFrame = frame;
-                frameBitmap.recycle();
+        Frame firstFrame = null;
+        Frame prevFrame = null;
+        Frame frame = null;
+        Bitmap frameBitmap;
+        for (int frameIdx = 0; frameIdx < animationFrameProvider.getFrameCount(); frameIdx++) {
```
lopeng: A list of bitmaps may or may not cause performance issue, for the same reason that a gif decoder won't decode all bitmaps in one shot, instead 2-3 when needed. Can we do stress test and measure the performance, so no more concern?

izzytwosheds: Each bitmap gets recycled once they are converted into a GL texture to free up memory.

lopeng: That's correct. But this design is assuming that all bitmaps need to be decoded then passed in. Think about a common case, showing gif, it getNextFrame/decodes one frame, show it, the recycle it, which has no concern. If you decode all then pass them, it might be a problem.

PR #10, opened 2020-01-14
Applying overlays when target video format is null

izzytwosheds:
In current implementation LiTr doesn't apply overlays if target video format is null. Which may be inconvenient for clients who want to modify video frames without changing resolution or bitrate. 

Now LiTr will make sure overlays are always applied. The catch is that on many devices bitrate is not provided for video tracks, for which we implement a utility method to estimate it.

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/MediaTransformer.java
+++ b/litr/src/main/java/com/linkedin/android/litr/MediaTransformer.java
@@ -55,6 +55,8 @@ public class MediaTransformer {
     public static final int GRANULARITY_NONE = 0;
     public static final int GRANULARITY_DEFAULT = 100;
 
+    public static final int DEFAULT_KEY_FRAME_INTERVAL = 5;
+
     private static final String TAG = MediaTransformer.class.getSimpleName();
     private static final int DEFAULT_FUTURE_MAP_SIZE = 10;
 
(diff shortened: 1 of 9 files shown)


litr/src/main/java/com/linkedin/android/litr/MediaTransformer.java
```diff
@@ -205,6 +214,29 @@ public void transform(@NonNull String requestId,
...
+                    .setTargetTrack(trackTransform.getTargetTrack())
+                    .setDecoder(trackTransform.getDecoder())
+                    .setEncoder(trackTransform.getEncoder())
+                    .setRenderer(trackTransform.getRenderer())
+                    .setTargetFormat(targetFormat)
+                    .build();
+
+                trackTransforms.remove(trackIndex);
+                trackTransforms.add(trackIndex, updatedTrackTransform);
+            }
```
lopeng: is that set()?

litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
```diff
@@ -109,6 +116,84 @@ public static long getEstimatedTargetVideoFileSize(@NonNull MediaSource mediaSou
         return getEstimatedTargetFileSize(trackTransforms);
     }
 
+    /**
+     * Estimates video track bitrate. On many devices bitrate value is not specified in {@link MediaFormat} for video track.
+     * Since not all data required for accurate estimation is available, this method makes several assumptions:
+     *  - if multiple video tracks are present, average per-pixel bitrate is assumed to be the same for all tracks
+     *  - if both bitrate and duration are not specified for a track, its size is not accounted for
```
lopeng: line 140 if both bitrate and duration; line 144 else, either bitrate or duration is not?

litr/src/main/java/com/linkedin/android/litr/MediaTransformer.java
izzytwosheds: Yes! Fixing...

litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
izzytwosheds: Good call, changing the wording

PR #18, opened 2020-04-17
Fixing MarshallingHandler not using provided looper

cerisier:
Provided looper was not used to initialize the MarshallingHandler which would use the `Looper.myLooper()` of whatever thread the transform job was submitted on.

I'm not sure how to test this nor if it makes sense to. But happy to follow your guidance if you have any idea :)

Best and thanks for this library 👍

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/MarshallingTransformationListener.java
+++ b/litr/src/main/java/com/linkedin/android/litr/MarshallingTransformationListener.java
@@ -46,7 +46,7 @@ class MarshallingTransformationListener {
         this.listener = listener;
 
         if (looper != null) {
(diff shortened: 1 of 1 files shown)


izzytwosheds:
Nice catch! 
Not sure how to test this, either. :)

izzytwosheds:
Also, thank you for the fix!

cyrilmottier:
Thanks @izzytwosheds. Are you going to do a new release soon?

izzytwosheds:
Just kicked it off, should be available in 10-15 minutes.

cyrilmottier:
Awesome. Thanks :)

PR #19, opened 2020-04-22
Support for custom video frame renderers

izzytwosheds:
GlVideoRenderer performed both management of input/output surfaces and rendering from input onto output surface. Rendering was very simple "scale to fit" between the two surfaces. 

Now splitting that logic into two: 
 - renderer manages surfaces
 - video frame rendering from input onto output surface is implemented as a filter. We introduce new type of filters to do this, which is given access to input video surface texture id and matrix.
 - default frame rendering filter implementation is current "scale to fit" mapping.

This refactor will allow us to add more sophisticated frame rendering filters implementing transformations like cropping, rotation, etc.

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/filter/video/gl/ScaleToFitGlFrameRenderFilter.java
+++ b/litr/src/main/java/com/linkedin/android/litr/filter/video/gl/ScaleToFitGlFrameRenderFilter.java
@@ -70,14 +70,14 @@ public class ScaleToFitGlFrameRenderFilter implements GlFrameRenderFilter {
     };
 
     private float[] mvpMatrix = new float[16];
-    private float[] inputTextureTransformMatrix = new float[16];
+    private float[] inputFrameTextureMatrix = new float[16];
 
     private int glProgram;
     private int mvpMatrixHandle;
     private int uStMatrixHandle;
     private int aPositionHandle;
     private int aTextureHandle;
-    private int inputTextureId;
+    private int inputFrameTextureHandle;
 
     @Override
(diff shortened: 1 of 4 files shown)


litr/src/main/java/com/linkedin/android/litr/filter/video/gl/ScaleToFitGlFrameRenderFilter.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private float[] mvpMatrix = new float[16];
+    private float[] inputTextureTransformMatrix = new float[16];
+
+    private int glProgram;
+    private int mvpMatrixHandle;
+    private int uStMatrixHandle;
+    private int aPositionHandle;
+    private int aTextureHandle;
+    private int inputTextureId;
```
lopeng, line 80: Keep the same coding style and sequence of the original open source might make later maintenance/upgrade easier. Thought?

izzytwosheds: Good call. Done.

lopeng: If check the original file https://android.googlesource.com/platform/cts/+/lollipop-release/tests/tests/media/src/android/media/cts/TextureRender.java, you will find mvpMatrixHandle was from muMVPMatrixHandle, uStMatrixHandle was from muSTMatrixHandle, etc. I am think if the two files can be diff easily, and patch any fix easily. But let you decide.

PR #22, opened 2020-04-23
Solid background color filter

izzytwosheds:


litr-filters/src/main/java/com/linkedin/android/litr/filter/video/gl/SolidBackgroundColorFilter.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Create using OpenGL color format
+     * @param red red value
+     * @param green green value
+     * @param blue blue value
+     */
+    public SolidBackgroundColorFilter(@FloatRange(from = 0, to = 1) float red,
+                                      @FloatRange(from = 0, to = 1) float green,
+                                      @FloatRange(from = 0, to = 1) float blue) {
```
lopeng, line 39: RGBA, Alpha? or later?

izzytwosheds: I couldn't come up with good use case for it. We are clearing canvas with that color, not rendering it. By default, canvas is black, so applying transparent background on top of black background will only darken it, which we can do ourselves by passing in lower intensity color values. And we don't have a use case for producing transparent video.
We can always add alpha channel support later.

PR #23, opened 2020-04-23
Filter for free transform video frame rendering

izzytwosheds:
A frame render filter, which supports geometric transformation of a source video frame when rendering it onto a target video frame:
 - since source and target video orientations no longer match, client must pass in target video orientation when using this renderer, if orientations don't match
 - scaling/rotating/positioning logic works pretty much the same way as for bitmap overlays. Except now we have to know orientation of a source video to display it correctly. 
 - new demo fragment allows experimenting with different setups: target video orientation, transform parameters, etc.

### DIFF
--- a/litr-filters/src/main/java/com/linkedin/android/litr/filter/video/gl/FreeTransformFrameRenderFilter.java
+++ b/litr-filters/src/main/java/com/linkedin/android/litr/filter/video/gl/FreeTransformFrameRenderFilter.java
@@ -1,0 +1,171 @@
+package com.linkedin.android.litr.filter.video.gl;
+
+import android.graphics.PointF;
+import android.opengl.GLES11Ext;
+import android.opengl.GLES20;
+import android.opengl.Matrix;
+
+import androidx.annotation.NonNull;
+
+import com.linkedin.android.litr.filter.GlFrameRenderFilter;
+import com.linkedin.android.litr.render.GlRenderUtils;
+
+/**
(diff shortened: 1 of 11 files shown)


litr-filters/src/main/java/com/linkedin/android/litr/filter/video/gl/FreeTransformFrameRenderFilter.java
```diff
@@ -0,0 +1,171 @@
...
+import androidx.annotation.NonNull;
+
+import com.linkedin.android.litr.filter.GlFrameRenderFilter;
+import com.linkedin.android.litr.render.GlRenderUtils;
+
+/**
+ * Implementation of {@link GlFrameRenderFilter} which can transform (scale, rotate, translate)
+ * source video frame when rendering it onto targetVideoFrame
+ */
+public class FreeTransformFrameRenderFilter implements GlFrameRenderFilter {
```
lopeng, line 17: A lot of repeating code from ScaleToFitGiFrameRenderFilter.java, make reviewing hard to focus on the key changes. Plan? I wait to review this file later.

izzytwosheds: I am planning to refactor all the code into common util methods later. I duplicated it for now to isolate new logic and make things easier to review. There are also a lot of similarities with bitmap overlay code, especially for MVP matrix computation.

PR #35, opened 2020-06-29
Implementation of PassthroughSoftwareRenderer

izzytwosheds:
Factoring out "rendering" logic in AudioTrackTranscoder into a separate PassthroughSoftwareRenderer, which simply copies decoder output buffer into encoder input buffer. Renderer is "synchronous" - when it is asked to render, it waits for encoder to dequeue an input buffer, then fills it with input frame data and releases back to encoder. If wait times out, buffer is dropped, similar to video GL renderer.

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
+++ b/litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2019 LinkedIn Corporation
+ * All Rights Reserved.
+ *
+ * Licensed under the BSD 2-Clause License (the "License").  See License in the project root for
+ * license information.
+ */
+package com.linkedin.android.litr.render;
+
+import android.media.MediaCodec;
+import android.media.MediaFormat;
+import android.util.Log;
+import android.view.Surface;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
+
+import com.linkedin.android.litr.codec.Encoder;
+import com.linkedin.android.litr.codec.Frame;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.TimeUnit;
+
+public class PassthroughSoftwareRenderer implements Renderer {
+
+    @VisibleForTesting static final long FRAME_WAIT_TIMEOUT = TimeUnit.SECONDS.toMicros(10);
+
+    private static final String TAG = "PassthroughSwRenderer";
+
+    @NonNull public final Encoder encoder;
+
+    public PassthroughSoftwareRenderer(@NonNull Encoder encoder) {
+        this.encoder = encoder;
+    }
+
+    @Override
+    public void init(@Nullable Surface outputSurface, @Nullable MediaFormat sourceMediaFormat, @Nullable MediaFormat targetMediaFormat) {}
+
+    @Nullable
+    @Override
+    public Surface getInputSurface() {
+        return null;
+    }
+
+    @Override
+    public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
+        if (frame == null || frame.buffer == null) {
+            Log.e(TAG, "Null or empty input frame provided");
+            return;
+        }
+        int tag = encoder.dequeueInputFrame(FRAME_WAIT_TIMEOUT);
+        if (tag >= 0) {
+            Frame outputFrame = encoder.getInputFrame(tag);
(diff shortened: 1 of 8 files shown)


litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -0,0 +1,89 @@
...
+        return null;
+    }
+
+    @Override
+    public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
+        if (frame == null || frame.buffer == null) {
+            Log.e(TAG, "Null or empty input frame provided");
+            return;
+        }
+        int tag = encoder.dequeueInputFrame(FRAME_WAIT_TIMEOUT);
```
lopeng, line 52: Any thoughts about how the default 10 sec timeout is selected? Any potential problem to be aware if the timeout is too long/short?

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            ByteBuffer inputBuffer = frame.buffer.asReadOnlyBuffer();
+            inputBuffer.rewind();
+            outputFrame.buffer.put(inputBuffer);
+
+            outputFrame.bufferInfo.set(
+                    0,
+                    frame.bufferInfo.size,
+                    frame.bufferInfo.presentationTimeUs,
```
lopeng, line 67: When to use the input arg presentationTimeNs?

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
izzytwosheds: I think this migrated from video track transcoder. I believe, the idea what that we would not wait indefinitely but long enough. My gut feel is that 10 seconds is long enough, MediaCodec works pretty fast, and at the same time 10 seconds is not too long in human perception. I think, in the future we can improve this logic to queue the audio frame instead of dropping it.

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
izzytwosheds: Here is doesn't matter much, I believe, since in passthrough mode we want to duplicate input buffer. That value is needed in GL renderer, which has common interface with software renderer.

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
lopeng: 10 sec seems very long for one frame. Just want to confirm if a long timeout (for some reason) will not cause dropping/impacting the following more frames.

izzytwosheds: I feel that relationship between timeout and frame drops is opposite - longer timeout should result in less chance of frame drop.

lopeng: Longer timeout results in less chance of **current** frame drop. If the transcoding need to be done in a limit time frame, longer timeout might cause more drops the following processing.

lopeng: I guess no such use case is supported, and the processing just takes longer to finish. So it should be fine here.

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
lopeng: Right. With passthrough in the class name, hope the client will not set presentationTimeUs then expect it to be honored.

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
izzytwosheds: I see your point. Realistically, I don't think that will be a big problem - if nothing happened in 10 seconds, chances are that things are already so bad, that transcoding will happen at all... Kinda like out of memory situation - when it happens, some small thing like a frame drop is the least of our problems.

lopeng: +

PR #36, opened 2020-07-01
allow to override fragment shader

tguerin:
For one of my project I had to apply a black and white filter to the video. To achieve this I had to extend the GlVideoRenderer to modify the fragment shader. Would be nice to be able to directly override the default fragment shader to apply some transformations upon GlVideoRenderer instantiation.


tguerin:
Sure, here it is https://github.com/linkedin/LiTr/pull/37

PR #37, opened 2020-07-01
add Grayscale shader

tguerin:
As requested, here is the shader to have a black and white filter. Used the same filter as https://github.com/MasayukiSuda/GPUVideo-android/blob/master/gpuv/src/main/java/com/daasuu/gpuv/egl/filter/GlGrayScaleFilter.java

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/shader/Shaders.java
+++ b/litr/src/main/java/com/linkedin/android/litr/shader/Shaders.java
@@ -1,0 +1,17 @@
+package com.linkedin.android.litr.shader;
+
+public class Shaders {
+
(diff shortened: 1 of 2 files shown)


litr/src/main/java/com/linkedin/android/litr/shader/Shaders.java
```diff
@@ -0,0 +1,17 @@
+package com.linkedin.android.litr.shader;
+
+public class Shaders {
```
izzytwosheds, line 3: Can we rename this to FragmentShaders?

tguerin: Sure

PR #56, opened 2020-09-14
Generalize drawable positioning into Transform data class

izzytwosheds:
Refactor repetitive size-position-rotation parameters into Transform data class. Using this class for bitmap overlays, as well as video frame positioning.


litr/src/main/java/com/linkedin/android/litr/filter/Transform.java
```diff
@@ -0,0 +1,31 @@
+package com.linkedin.android.litr.filter;
```
lopeng, line 1: copyright?

izzytwosheds: Good call. Fixing...

lopeng (approved):

PR #81, opened 2021-01-13
Take MediaRange into account when estimating target size

niekdev:
Whenever the library estimates the target media's size, it always checks for the track durations, not the user defined `MediaRange`s. This PR adds the missing functionality which will improve the size estimation when video trimming is involved.

PR includes a test where the trimmed (half the mocked track length) estimated size is half of the default track size.


izzytwosheds (approved):
Well done! Thank for for doing this.

Issue #102, opened 2021-03-08
Failed to transcode some video files


Ma7moudHatem:
I have the following video file, which I need to transcode it by either add an overlay above it or change the bitrate of the video,
https://drive.google.com/file/d/1Y6t1jIchWbNl_f71MA1mq56jrKjxgHTs/view 
 I tried the demo app, and when clicking on the transcode button, it keeps processing, and after I debugged the issue, I found that `LiTr` will throw a `BufferOverflowException`, and when it tries to release the `MediaMuxerMediaTarget`, it also throws an exception in the `release()` method, so the `listeners` will not receive the stop status with the failed state.

NOTES:
1. I'm currently using` litr 1.4.9`
2. I transcode the video to `video/avc` format, with a video bitrate 1.5Mb/s
3. I transcode the audio to `audio/mp4a-latm` format, with an audio bitrate 128Kb/s
4. I think the issue related to the **audio track**, so when just transcoding the audio track, the issue happened.

What I need:
 1. Please if you can fix the transcoding issue related to the attached video file.
 2. Make sure that if an exception happened in the `release(boolean success)` method inside `TransformationJob` class will be caught and handled.

izzytwosheds:
Thank you for reporting! I will look into this.

izzytwosheds:
Which device/OS version are you using? I cannot reproduce it on Samsung Note 10. 
Can you also post the logcat?

Ma7moudHatem:
As I mentioned before, the issue happened only if I transcoded the audio track, and after some debugging, I found the line of code that causes the issue, In `renderFrame` method inside `PassthroughSoftwareRenderer` class, at line 62  `outputFrame.buffer.put(inputBuffer);` , an exception(`BufferOverflowException`) will be thrown, because somehow the `outputFrame buffer`'s remaining is less than the `inputBuffer` remaining, for example, the outputFrame buffer remaining is **4096** and the inputBuffer remaining is **12250**.

Ma7moudHatem:
@izzytwosheds please, when you will release this pull request https://github.com/linkedin/LiTr/pull/103 ? which will fix the current issue.

izzytwosheds:
Released it, should be available now.

PR #103, opened 2021-03-09
Fix renderFrame logic in PassthroughSoftwareRenderer

Ma7moudHatem:
1. Fix an issue in renderFrame() method inside PassthroughSoftwareRenderer class, where it throws a BufferOverflowException when the InputBuffer's size is larger than the OutputBuffer's size.
2. Add a try-catch to some release() methods, where it's safe to catch the exception and not throw it, also to not make wrong behavior where the onError method of TransformationListener will not be called.
3. This pull request will fix the opened issue in litr, https://github.com/linkedin/LiTr/issues/102

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/TransformationJob.java
+++ b/litr/src/main/java/com/linkedin/android/litr/TransformationJob.java
@@ -245,10 +245,20 @@ class TransformationJob implements Runnable {
             mediaTargets.add(trackTransform.getMediaTarget());
         }
         for (MediaSource mediaSource : mediaSources) {
-            mediaSource.release();
+            try {
+                // Sometimes this method will throw an exception, so we just need to catch it.
+                mediaSource.release();
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
         }
         for (MediaTarget mediaTarget : mediaTargets) {
-            mediaTarget.release();
+            try {
+                // Sometimes this method will throw an exception, so we just need to catch it.
+                mediaTarget.release();
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
             if (!success) {
                 deleteOutputFile(mediaTarget.getOutputFilePath());
             }
--- a/litr/src/main/java/com/linkedin/android/litr/TransformationJob.java
+++ b/litr/src/main/java/com/linkedin/android/litr/TransformationJob.java
@@ -245,20 +245,10 @@ class TransformationJob implements Runnable {
             mediaTargets.add(trackTransform.getMediaTarget());
         }
         for (MediaSource mediaSource : mediaSources) {
-            try {
-                // Sometimes this method will throw an exception, so we just need to catch it.
-                mediaSource.release();
-            } catch (Exception e) {
-                e.printStackTrace();
-            }
+            mediaSource.release();
         }
         for (MediaTarget mediaTarget : mediaTargets) {
-            try {
-                // Sometimes this method will throw an exception, so we just need to catch it.
-                mediaTarget.release();
-            } catch (Exception e) {
-                e.printStackTrace();
-            }
+            mediaTarget.release();
             if (!success) {
                 deleteOutputFile(mediaTarget.getOutputFilePath());
             }
--- a/litr/src/main/java/com/linkedin/android/litr/codec/MediaCodecDecoder.java
+++ b/litr/src/main/java/com/linkedin/android/litr/codec/MediaCodecDecoder.java
@@ -105,7 +105,7 @@ public final class MediaCodecDecoder implements Decoder {
 
     @Override
     public int dequeueOutputFrame(long timeout) {
-        return mediaCodec.dequeueOutputBuffer(outputBufferInfo, 0);
+        return mediaCodec.dequeueOutputBuffer(outputBufferInfo, timeout);
     }
 
     @Override
--- a/litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
+++ b/litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
@@ -20,6 +20,7 @@ import com.linkedin.android.litr.codec.Encoder;
 import com.linkedin.android.litr.codec.Frame;
 
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
 import java.util.concurrent.TimeUnit;
 
 public class PassthroughSoftwareRenderer implements Renderer {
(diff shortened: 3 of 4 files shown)


litr/src/main/java/com/linkedin/android/litr/TransformationJob.java
```diff
@@ -245,10 +245,20 @@ void release(boolean success) {
             mediaTargets.add(trackTransform.getMediaTarget());
         }
         for (MediaSource mediaSource : mediaSources) {
-            mediaSource.release();
+            try {
+                // Sometimes this method will throw an exception, so we just need to catch it.
+                mediaSource.release();
```
izzytwosheds, line 250: So I presume that if we catch this exception transformation completes and resulting transcoded video/audio is playable? 
Did you check what root cause of this might be?

litr/src/main/java/com/linkedin/android/litr/codec/MediaCodecDecoder.java
```diff
@@ -105,7 +105,7 @@ public void queueInputFrame(@NonNull Frame frame) {
 
     @Override
     public int dequeueOutputFrame(long timeout) {
-        return mediaCodec.dequeueOutputBuffer(outputBufferInfo, 0);
+        return mediaCodec.dequeueOutputBuffer(outputBufferInfo, timeout);
```
izzytwosheds, line 108: nice catch, thank you!

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -49,33 +50,56 @@ public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
...
-                Log.e(TAG, "No input frame returned by an encoder, dropping a frame");
-                return;
-            }
 
-            ByteBuffer inputBuffer = frame.buffer.asReadOnlyBuffer();
-            inputBuffer.rewind();
-            outputFrame.buffer.put(inputBuffer);
+        ArrayList<Frame> outputFrames = new ArrayList<>();
+        ByteBuffer inputBuffer = null;
+        boolean isFirst = true;
```
izzytwosheds, line 56: rename to `isNewInputFrame`, to make it more readable

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -49,33 +50,56 @@ public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
...
+        boolean isContinue;
```
izzytwosheds, line 57: rename to `areBytesRemaining`, to make it more readable

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -49,33 +50,56 @@ public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
...
+            }
+        }
+        while (isContinue);
 
-            outputFrame.bufferInfo.set(
-                    0,
-                    frame.bufferInfo.size,
-                    TimeUnit.NANOSECONDS.toMicros(presentationTimeNs),
-                    frame.bufferInfo.flags);
+        for (Frame outputFrame : outputFrames) {
```
izzytwosheds, line 101: Why do we need a separate loop for this? Can't we do this in the `do while` loop and enqueue encoder input frame as soon as we fill it with data?

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -49,33 +50,56 @@ public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
...
+                }
+
+                if (isFirst) {
+                    isFirst = false;
+                    inputBuffer = frame.buffer.asReadOnlyBuffer();
+                    inputBuffer.rewind();
+                }
+
+                int capacity = Math.min(outputFrame.buffer.remaining(), inputBuffer.remaining());
+                byte[] inputBytes = new byte[capacity];
```
izzytwosheds, line 75: Allocating new buffer for every frame will hurt performance (memory fragmentation, frequent GC invocation, etc.) Let's rename this to `inputByteBuffer`, make it a member variable and create a new one only when `capacity > inputByteBuffer.length`

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -49,33 +50,56 @@ public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
...
+                    return;
+                }
+
+                if (isFirst) {
+                    isFirst = false;
+                    inputBuffer = frame.buffer.asReadOnlyBuffer();
+                    inputBuffer.rewind();
+                }
+
+                int capacity = Math.min(outputFrame.buffer.remaining(), inputBuffer.remaining());
```
izzytwosheds, line 74: Another way is to have an `if (outputFrame.buffer.remaining() < inputBuffer.remaining()` comparison here and do buffer copy (new logic) only when it is true. Otherwise, we can just copy directly, like we were doing before.

izzytwosheds (commented):
Please add unit tests for new logic in `PassthroughSoftwareRendererShould`, it should be very straightforward.

litr/src/main/java/com/linkedin/android/litr/TransformationJob.java
Ma7moudHatem: In my case, when the InputBuffer's size is larger than the OutputBuffer's size, and try to write this buffer to the OutputBuffer, then this will throw a `BufferOverflowException`, and this exception will be caught in TransformationJob in the `run()` method, but when trying to call the `error()` method then `release()` method, the `mediaTarget.release()` will throw an exception but not caught.

Ma7moudHatem: Ok, great, currently I'll remove this change from the pull request, and let you make the proper fix for any exception that could be thrown from `mediaTarget.release()`

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -49,34 +51,58 @@ public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
...
+
+                if (isNewInputFrame) {
+                    isNewInputFrame = false;
+                    inputBuffer = frame.buffer.asReadOnlyBuffer();
+                    inputBuffer.rewind();
+                }
+
+                if (outputFrame.buffer.remaining() < inputBuffer.remaining()) {
+                    capacity = outputFrame.buffer.remaining();
+                    inputByteBuffer = new byte[capacity];
```
izzytwosheds, line 77: one last thing. You need to create inputByteBuffer only if it is too small or null, so surround this with 
`if (inputByteBuffer == null || inputByteBuffer.length < capacity)`

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -49,34 +51,58 @@ public void renderFrame(@Nullable Frame frame, long presentationTimeNs) {
...
+                    inputBuffer.get(inputByteBuffer, 0, capacity);
+
+                    outputFrame.buffer.put(inputByteBuffer);
```
izzytwosheds, line 80: here, you will have to use `outputFrame.buffer.put(inputByteBuffer, 0, capacity);`

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
Ma7moudHatem: Fixed

PR #104, opened 2021-03-16
- Fix an issue where the duration is less than 1 sec, which will divide by zero exception 

Ma7moudHatem:
- Fix an issue where the duration is less than 1 sec, which will cause a divide by zero exception.
- Convert the Micros to seconds as a float to count the fractions of a second

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
+++ b/litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
@@ -131,6 +131,10 @@ public final class TranscoderUtils {
         if (videoTrackFormat.containsKey(MediaFormat.KEY_BIT_RATE)) {
             return videoTrackFormat.getInteger(MediaFormat.KEY_BIT_RATE);
         }
+        float videoTrackDuration = TimeUtils.microsToSeconds(videoTrackFormat.getLong(MediaFormat.KEY_DURATION));
+        if (videoTrackDuration == 0) {
+            return 0;
+        }
 
         float unallocatedSize = mediaSource.getSize();
         float totalPixels = 0;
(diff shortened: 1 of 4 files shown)


izzytwosheds (commented):
Looks like you have three unrelated fixes in the same PR. Can you split them into separate PRs? This way we will have cleaner commit history.

litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
```diff
@@ -154,13 +154,15 @@ public static int estimateVideoTrackBitrate(@NonNull MediaSource mediaSource, in
             }
         }
 
+        long videoTrackDuration = TimeUnit.MICROSECONDS.toSeconds(videoTrackFormat.getLong(MediaFormat.KEY_DURATION));
```
izzytwosheds, line 157: I think it would be better to convert this to a float to handle fractions of a second. That would make our calculations more precise, too.

litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
```diff
@@ -133,34 +132,35 @@ public static int estimateVideoTrackBitrate(@NonNull MediaSource mediaSource, in
...
+        float trackPixels = videoTrackFormat.getInteger(MediaFormat.KEY_WIDTH)
             * videoTrackFormat.getInteger(MediaFormat.KEY_HEIGHT)
-            * TimeUnit.MICROSECONDS.toSeconds(videoTrackFormat.getLong(MediaFormat.KEY_DURATION));
+            * videoTrackDuration;
 
-        long trackSize = unallocatedSize * trackPixels / totalPixels;
-
-        return (int) (trackSize * 8 / TimeUnit.MICROSECONDS.toSeconds(videoTrackFormat.getLong(MediaFormat.KEY_DURATION)));
+        float trackSize = totalPixels > 0 ? unallocatedSize * trackPixels / totalPixels : unallocatedSize;
+        videoTrackDuration = videoTrackDuration == 0 ? 1 : videoTrackDuration;
```
izzytwosheds, line 162: I don't think this really applies anymore, since we are working with <1 fractional values now. Let's guard in a different way: 
// zero duration means empty track
if (videoTrackDuration == 0) {
    return 0; 
}

we can actually do this early and return early if duration is zero, since this means that track has no pixels in it.

Ma7moudHatem: Fixed

PR #109, opened 2021-03-25
Enhance the logic for finding a media codec for a media format

Ma7moudHatem:
Enhance the logic for finding a media codec for a media format, where sometimes the findEncoderForFormat/findDeccoderForFormat or createEncoderByType/createDecoderByType methods will return the first MediaCodec in the available media codecs that could support the media type, but this codec may not be able to configure the media format, so the fix here is to get all available codecs that support the mimeType, and iterate over them to check which codec could configure the media format and return it

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/utils/CodecUtils.java
+++ b/litr/src/main/java/com/linkedin/android/litr/utils/CodecUtils.java
@@ -22,8 +22,11 @@ import androidx.annotation.RequiresApi;
 import com.linkedin.android.litr.exception.TrackTranscoderException;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
 
 public class CodecUtils {
 
@@ -172,9 +175,9 @@ public class CodecUtils {
      * @param codecFormatNotFoundError message to provide in {@link TrackTranscoderException} if codec could not be found by format
      * @param codecConfigurationError message to provide in {@link TrackTranscoderException} if codec could not configured
      * @param fallbackToGetCodecByType flag indicating if older getCodecByType API should be used if getting by format fails
-     * @return configured instance of {@link MediaCodec}, null if codec could not be found or configured
+     * @return configured instance of {@link MediaCodec}, or a {@link TrackTranscoderException} will be thrown
      */
-    @Nullable
+    @NonNull
     public static MediaCodec getAndConfigureCodec(@NonNull MediaFormat mediaFormat,
                                                   @Nullable Surface surface,
                                                   boolean isEncoder,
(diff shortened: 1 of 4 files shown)


litr/src/main/java/com/linkedin/android/litr/utils/CodecUtils.java
```diff
@@ -172,99 +175,129 @@ public static int getHighestSupportedProfile(@NonNull String mimeType, boolean i
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            MediaCodecList mediaCodecList = new MediaCodecList(MediaCodecList.ALL_CODECS);
+            for (MediaCodecInfo info : mediaCodecList.getCodecInfos()) {
+                if (info.isEncoder() != encoder) {
+                    continue;
+                }
+                try {
+                    MediaCodecInfo.CodecCapabilities caps = info.getCapabilitiesForType(mimeType);
+                    if (caps != null && (mediaFormat == null || caps.isFormatSupported(mediaFormat))) {
+                        supportedMediaCodecs.add(() -> MediaCodec.createByCodecName(info.getName()));
```
izzytwosheds, line 288: Just curious, why are you using `Callable`?

Ma7moudHatem: I think it could be better to not create multiple instances of `Mediacodec`, where most of the time I'll use the first one, and I'll need to releases all other created instances that unneeded. so I used the `Callable` to create `MediaCodec` instance when needed only.

litr/src/main/java/com/linkedin/android/litr/utils/CodecUtils.java
```diff
@@ -172,99 +175,129 @@ public static int getHighestSupportedProfile(@NonNull String mimeType, boolean i
...
-        String codecName = isEncoder
-                ? mediaCodecList.findEncoderForFormat(mediaFormat)
-                : mediaCodecList.findDecoderForFormat(mediaFormat);
-        if (codecName != null) {
-            mediaCodec = MediaCodec.createByCodecName(codecName);
-        }
-        if (mediaCodec != null) {
-            mediaCodec.configure(mediaFormat, surface, null, isEncoder ? MediaCodec.CONFIGURE_FLAG_ENCODE : 0);
+    @NonNull
+    private static MediaCodec getAndConfigureCodec(
```
izzytwosheds, line 231: Rename this to `createAndConfigureCodec`, more accurate to what method does

litr/src/main/java/com/linkedin/android/litr/utils/CodecUtils.java
```diff
@@ -172,99 +175,129 @@ public static int getHighestSupportedProfile(@NonNull String mimeType, boolean i
...
                                                   boolean isEncoder,
                                                   @NonNull TrackTranscoderException.Error codecNotFoundError,
                                                   @NonNull TrackTranscoderException.Error codecFormatNotFoundError,
                                                   @NonNull TrackTranscoderException.Error codecConfigurationError,
                                                   boolean fallbackToGetCodecByType) throws TrackTranscoderException {
-        MediaCodec mediaCodec = null;
         try {
-            if (Build.VERSION.SDK_INT > Build.VERSION_CODES.LOLLIPOP) {
-                mediaCodec = getAndConfigureCodecByConfig(mediaFormat, surface, isEncoder);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
izzytwosheds, line 189: From the functionality perspective this is a pretty big and important change. I usually do these in a way that allows A/B testing. For example, here I would recommend making your entire new logic a new set of methods and calling the new equivalent of this method from `MediaCodecEncoder` and `MediaCodecDecoder` depending on a flag we pass in in the constructor, similar to what we do with `fallbackToGetCodecByType`

litr/src/main/java/com/linkedin/android/litr/utils/CodecUtils.java
izzytwosheds: Very clever. :)

litr/src/main/java/com/linkedin/android/litr/utils/CodecUtils.java
Ma7moudHatem: @izzytwosheds could you please suggest a name for this flag?

izzytwosheds: filterByTypeAndFormat? lookupByTypeAndFormat?

Ma7moudHatem: Done

izzytwosheds (approved):

PR #111, opened 2021-03-31
Fix an issue in Audio transcoding where target-sample-rate is less than the source-sample-rate

Ma7moudHatem:
Fix an issue in Audio transcoding, where the target sample-rate is less than the source sample-rate, so here we need to do a down-sampling, and I found this fix by another third party, but I did some changes.


litr/src/main/java/com/linkedin/android/litr/render/Renderer.java
```diff
@@ -32,12 +32,15 @@
...
-     * @param frame {@link Frame} to operate with. Non-null ror non OpenGL renderer, will contain raw pixels.
+     * Render an input frame
+     * @param inputFrame {@link Frame} to operate with. Non-null ror non OpenGL renderer, will contain raw pixels.
      *                           null for GL renderer, which should assume that environment has been set and just invoke Gl calls.
      * @param presentationTimeNs frame presentation time in nanoseconds
+     * @param sourceMediaFormat the raw(source) {@link MediaFormat}
+     * @param targetMediaFormat the target {@link MediaFormat}
      */
-    void renderFrame(@Nullable Frame frame, long presentationTimeNs);
+    void renderFrame(@Nullable Frame inputFrame, long presentationTimeNs, MediaFormat sourceMediaFormat,
```
izzytwosheds, line 42: Why do we have to break a public API? Source and target formats are passed in `init` method, can't we just hold on to them in the renderer? They don't usually change during transcoding.

litr/src/main/java/com/linkedin/android/litr/resample/UpsampleAudioResampler.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * An {@link AudioResampler} that upsamples from a lower sample rate to a higher sample rate.
+ */
+public class UpsampleAudioResampler extends PassThroughAudioResampler {
+    @Override
+    public void resample(@NonNull ByteBuffer inputBuffer, int inputSampleRate, @NonNull ByteBuffer outputBuffer,
+            int outputSampleRate, int channels) {
+        super.resample(inputBuffer, inputSampleRate, outputBuffer, outputSampleRate, channels);
+        // TODO currently up-sampling not supported because it's need some enhancements, so fallback to PASSTHROUGH approach
```
izzytwosheds, line 14: We probably don't want to do this here. I think we can remove this entire class for now and bring it in later.

litr/src/main/java/com/linkedin/android/litr/resample/DownsampleAudioResampler.java
```diff
@@ -0,0 +1,49 @@
+package com.linkedin.android.litr.resample;
+
+import androidx.annotation.NonNull;
+import java.nio.ByteBuffer;
+
+/**
+ * An {@link AudioResampler} that downsamples from a higher sample rate to a lower sample rate.
+ */
+public class DownsampleAudioResampler implements AudioResampler {
```
izzytwosheds, line 9: Is this adapted code from a different open source project? In that case we have to keep their copyright notice and add a note that we made modifications (feel free to add your name as author of modifications)

litr/src/main/java/com/linkedin/android/litr/resample/AudioResampler.java
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Resamples input audio from input buffer into the output buffer.
+     *
+     * @param inputBuffer the input buffer
+     * @param inputSampleRate the input sample rate
+     * @param outputBuffer the output buffer
+     * @param outputSampleRate the output sample rate
+     * @param channels the number of channels
+     */
+    void resample(@NonNull final ByteBuffer inputBuffer, int inputSampleRate, @NonNull final ByteBuffer outputBuffer, int outputSampleRate, int channels);
```
izzytwosheds, line 21: I think, factory pattern would work better here to create correct resampler implementation for given source/target formats. We would need instantiate only one class.

litr/src/main/java/com/linkedin/android/litr/resample/AudioResampler.java
```diff
@@ -0,0 +1,28 @@
+package com.linkedin.android.litr.resample;
+
+import androidx.annotation.NonNull;
+import java.nio.ByteBuffer;
+
+/**
+ * Resamples audio data. See {@link UpsampleAudioResampler} or
+ * {@link DownsampleAudioResampler} for concrete implementations.
+ */
+public interface AudioResampler {
```
izzytwosheds, line 10: For all new classes that are not modified classes from other project please add LinkedIn copyright notice

litr/src/main/java/com/linkedin/android/litr/render/Renderer.java
Ma7moudHatem: No, we don't need to break the `renderFrame()` method, but yes, they change after `init` method, so I'll add a new method `onMediaFormatChanged()` to `Renderer` interface, and will call it each time the `MediaCodec.INFO_OUTPUT_FORMAT_CHANGED` is returned

litr/src/main/java/com/linkedin/android/litr/resample/UpsampleAudioResampler.java
Ma7moudHatem: ok

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -31,7 +36,20 @@
...
     public final long frameWaitTimeoutUs;
 
-    private byte[] inputByteBuffer = null;
+    private MediaFormat sourceAudioFormat;
+    private MediaFormat targetAudioFormat;
+
+    private final AudioResampler audioResampler = new DefaultAudioResampler();
+
+    @VisibleForTesting
+    static int getSampleRate(MediaFormat format) {
```
izzytwosheds, line 45: Why is this static?

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -31,7 +36,20 @@
...
+        return format.getInteger(MediaFormat.KEY_SAMPLE_RATE);
+    }
+
+    @VisibleForTesting
+    static int getChannels(MediaFormat format) {
```
izzytwosheds, line 50: And this?

litr/src/main/java/com/linkedin/android/litr/resample/AudioResampler.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Resamples input audio from input buffer into the output buffer.
+     *
+     * @param inputBuffer the input buffer
+     * @param inputSampleRate the input sample rate
+     * @param outputBuffer the output buffer
+     * @param outputSampleRate the output sample rate
+     * @param channels the number of channels
+     */
+    void resample(@NonNull final ByteBuffer inputBuffer, int inputSampleRate, @NonNull final ByteBuffer outputBuffer, int outputSampleRate, int channels);
```
izzytwosheds, line 37: I think it would be more flexible to pass `MediaFormat` for source and target. Resampler implementation would know how to extract information it needs from it. And we would not need to change method signature (breaking change) if we needed to pass in more parameters than we do now.

litr/src/main/java/com/linkedin/android/litr/resample/AudioResampler.java
```diff
@@ -0,0 +1,42 @@
...
+
+    AudioResampler DOWNSAMPLE = new DownsampleAudioResampler();
```
izzytwosheds, line 39: Interface being aware of two of its implementations does not sounds right. We should use factory pattern to instantiate a correct resampler implementation in `onMediaFormatChanged` method. By default, it should be `PassthroughResampler`, of course.

litr/src/main/java/com/linkedin/android/litr/resample/DefaultAudioResampler.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * An {@link AudioResampler} that delegates to appropriate classes
+ * based on input and output size.
+ */
+public class DefaultAudioResampler implements AudioResampler {
+
+    @Override
+    public void resample(@NonNull ByteBuffer inputBuffer, int inputSampleRate, @NonNull ByteBuffer outputBuffer, int outputSampleRate, int channels) {
+        if (inputSampleRate > outputSampleRate) {
```
izzytwosheds, line 31: This logic can move into `onMediaFormatChanged` method. That way we would remove this entire class and only one instance of resampler would be created. This would also remove those two constants with implementations from `AudioResampler` interface.

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
Ma7moudHatem: changed to non-static method

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
Ma7moudHatem: removed from here

litr/src/main/java/com/linkedin/android/litr/resample/DownsampleAudioResampler.java
```diff
@@ -0,0 +1,73 @@
...
+import android.media.MediaFormat;
+import androidx.annotation.NonNull;
+import java.nio.ByteBuffer;
+
+/**
+ * An {@link AudioResampler} that downsamples from a higher sample rate to a lower sample rate.
+ */
+public class DownsampleAudioResampler implements AudioResampler {
+
+    private static float ratio(int remaining, int all) {
```
izzytwosheds, line 31: I guess this was original static and got copied there? Can we make this not static?

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -42,8 +50,35 @@ public PassthroughSoftwareRenderer(@NonNull Encoder encoder, long frameWaitTimeo
         this.frameWaitTimeoutUs = frameWaitTimeoutUs;
     }
 
+    private int getSampleRate(MediaFormat format) {
```
izzytwosheds, line 53: Do we need this method?

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
```diff
@@ -42,8 +50,35 @@ public PassthroughSoftwareRenderer(@NonNull Encoder encoder, long frameWaitTimeo
...
+        this.sourceAudioFormat = sourceMediaFormat;
+        this.targetAudioFormat = targetMediaFormat;
+        initAudioResampler();
+    }
+
+    private void initAudioResampler() {
+        if (sourceAudioFormat == null || targetAudioFormat == null) {
+            return;
+        }
+        int inputSampleRate = getSampleRate(sourceAudioFormat);
```
izzytwosheds, line 74: Can't we use `AudioResampler` method here? It seems to be doing the same thing.

Ma7moudHatem: done

litr/src/main/java/com/linkedin/android/litr/render/PassthroughSoftwareRenderer.java
Ma7moudHatem: Not anymore

izzytwosheds (approved):

PR #114, opened 2021-04-20
Add support to write data to a File using a Uri

Ma7moudHatem:
Add support to write data to a File using a Uri, where most likely to support scoped-storage enforcement in Android 10+, by using Uri instead of directly writing to File which will cause an exception.


litr/src/main/java/com/linkedin/android/litr/io/MediaMuxerMediaTarget.java
```diff
@@ -40,16 +45,51 @@
...
 
+    private ParcelFileDescriptor parcelFileDescriptor;
     private String outputFilePath;
     private int numberOfTracksToAdd;
     private int trackCount;
 
+    /**
+     * This constructor to support scoped-storage enforcement in Android 10+, by using Uri instead of directly writing to File
+     */
+    public MediaMuxerMediaTarget(@NonNull Context context, @NonNull Uri outputFileUri,
```
izzytwosheds, line 56: I like the idea of this new constructor. But I feel its logic might be overcomplicated. What it should do is this:
 - if API level is >= 26 it should use FileDescriptor when creating a MediaMuxer
 - at lower API levels it should behave exactly the way current constructor does - accept only "file://"  URI
 - otherwise (non-file URI in API level < 26) we should throw a `MediaTargetException`, maybe add a new error type in it for UNSUPPORTED_URI_TYPE with "URI type not supported at API level below 26" description.
This way, a constructor can create a `MediaMuxer` and pass it to `init` method to be configured.

litr/src/main/java/com/linkedin/android/litr/io/MediaMuxerMediaTarget.java
```diff
@@ -40,16 +45,51 @@
 
     private MediaFormat[] mediaFormatsToAdd;
 
+    private ParcelFileDescriptor parcelFileDescriptor;
     private String outputFilePath;
     private int numberOfTracksToAdd;
     private int trackCount;
 
+    /**
+     * This constructor to support scoped-storage enforcement in Android 10+, by using Uri instead of directly writing to File
```
izzytwosheds, line 54: Documentation should change to:
Create an instance using input URI. On Android Oreo (API level 26) and above it can be any URI writeable by the provided context, otherwise it must be a "file://" URI. This constructor is especially useful when working with FileProvider created URI to comply with scoped storage enforcement on Android 10+

litr/src/main/java/com/linkedin/android/litr/io/MediaMuxerMediaTarget.java
```diff
@@ -40,16 +45,51 @@
...
+            releaseFileDescriptor();
+            throw new MediaTargetException(IO_FAILUE, outputFilePath, outputFormat, ioException);
+        }
+    }
+
     public MediaMuxerMediaTarget(@NonNull String outputFilePath, @IntRange(from = 1) int trackCount, int orientationHint, int outputFormat) throws MediaTargetException {
+        init(outputFilePath, trackCount, orientationHint, outputFormat);
+    }
+
+    private void init(@NonNull String outputFilePath, @IntRange(from = 1) int trackCount, int orientationHint,
```
izzytwosheds, line 84: Pass `MediaMuxer` instead of `outputFilePath`

litr/src/main/java/com/linkedin/android/litr/utils/FileUtils.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Get a file path from a Uri. This will get the the path for Storage Access
+     * Framework Documents, as well as the _data field for the MediaStore and
+     * other file-based ContentProviders.
+     *
+     * Callers should check whether the path is local before assuming it represents a local file.
+     */
+    @Nullable
+    public static String getFilePathFromUri(Context context, Uri uri){
```
izzytwosheds, line 36: We will not need this.

litr/src/main/java/com/linkedin/android/litr/utils/FileUtils.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.FileNotFoundException;
+
+public class FileUtils {
+
+    private FileUtils() {}
+
+    /**
+     * Open a raw file descriptor to access data under a URI.
+     */
+    public static ParcelFileDescriptor getParcelFileDescriptor(Context context, Uri outputFileUri) throws FileNotFoundException {
```
izzytwosheds, line 24: Technically, after suggested changes this entire class can go away and become a single method call in `MediaMuxerMediaTarget`

litr/src/main/java/com/linkedin/android/litr/io/MediaMuxerMediaTarget.java
Ma7moudHatem: Fixed

litr/src/main/java/com/linkedin/android/litr/io/MediaMuxerMediaTarget.java
```diff
@@ -40,29 +44,66 @@
...
         try {
-            mediaMuxer = new MediaMuxer(outputFilePath,  outputFormat);
-            mediaMuxer.setOrientationHint(orientationHint);
+            MediaMuxer mediaMuxer;
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                parcelFileDescriptor = context.getContentResolver().openFileDescriptor(outputFileUri, "rwt");
+                if (parcelFileDescriptor != null) {
+                    mediaMuxer = new MediaMuxer(parcelFileDescriptor.getFileDescriptor(), outputFormat);
+                } else {
+                    throw new IOException();
```
izzytwosheds, line 66: Maybe MediaTargetException here as well? Use IO_FAILURE enum value

Ma7moudHatem: @izzytwosheds, already the `MediaTargetException` with `IO_FAILURE` will be thrown after throwing the `IOException` because there is an `IOException` catch case with the try, else if you prefer to directly throw the `MediaTargetException` here.

izzytwosheds: Gotcha, I see it now. Should we add a helpful message to IOException? Something like "Inaccessible URI" + outputFileUri ?

Ma7moudHatem: I think yes, and it's fixed now

izzytwosheds (approved):

PR #135, opened 2021-09-29
Fix GlVideoRenderer.hasFilters() logic

izzytwosheds:
`GlVideoRenderer.hasFilters()` always returned true, since it always had a `DefaultFrameRendererFilter` 
Intended behavior of `hasFilters()` method is to see if renderer has user provided filters, which would mean that user is modifying video frames, which in turn means that we cannot use a `PassthroughTrackTranscoder`. Fixing this now.


ReallyVasiliy (approved):
👍 
We should be mindful of updating the flag if we provide a way to modify `filters` outside the ctor in the future

PR #145, opened 2021-11-02
Support for recording video frames

ReallyVasiliy:
TODO


izzytwosheds:
There is a working implementation now. Closing this, it will still be available for reference.

PR #146, opened 2021-11-09
Support for extracting a series of video thumbnails

ReallyVasiliy:
This PR adds support for extracting thumbnails/frames from video as `Bitmap`s, while applying an optional set of filters.

`VideoThumbnailExtractor` serves as the entry point, and requires `ThumbnailExtractParameters` which defines the extraction behavior and other parameters.

One behavior, `MediaMetadataExtractionBehavior` (based on `MediaMetadataRetriever`), is included. The intent is that other behaviors can be added by consumers -- for example, consumers might opt to provide a custom FFmpeg-based or MediaCodec-based extractor, which may yield performance or functional improvements over `MediaMetadataRetriever`, while still retaining the ability to apply video filters to extracted frames.

Demo app has an example of creating a "film strip":

Screenshot_1636489276


litr/src/main/java/com/linkedin/android/litr/thumbnails/behaviors/MediaMetadataExtractionBehavior.kt
```diff
@@ -0,0 +1,61 @@
...
+            }
+            ExtractionMode.ExactFrameOnly -> {
+                completed = retrieveThumbnails(params, MediaMetadataRetriever.OPTION_CLOSEST, listener)
+            }
+            ExtractionMode.TwoPass -> {
+                completed = retrieveThumbnails(params, MediaMetadataRetriever.OPTION_CLOSEST_SYNC, listener)
+                if (completed) {
+                    completed = retrieveThumbnails(params, MediaMetadataRetriever.OPTION_CLOSEST, listener)
+                }
+            }
```
ReallyVasiliy, line 24: No need for two-pass: client can run extract twice

litr/src/main/java/com/linkedin/android/litr/thumbnails/ThumbnailExtractParameters.kt
```diff
@@ -0,0 +1,29 @@
+package com.linkedin.android.litr.thumbnails
+
+import android.graphics.Point
+import com.linkedin.android.litr.render.ThumbnailRenderer
+import com.linkedin.android.litr.thumbnails.behaviors.ExtractionBehavior
+
+/**
+ * Request parameters for thumbnail extraction.
+ */
+data class ThumbnailExtractParameters(
```
ReallyVasiliy, line 10: Add an option to write out the thumbnails to a file, and return file handle

litr/src/main/java/com/linkedin/android/litr/thumbnails/behaviors/ExtractionBehavior.kt
```diff
@@ -0,0 +1,38 @@
+package com.linkedin.android.litr.thumbnails.behaviors
```
ReallyVasiliy, line 1: Add Copyright notice to all new files

litr/src/main/java/com/linkedin/android/litr/thumbnails/VideoThumbnailExtractor.kt
```diff
@@ -0,0 +1,89 @@
...
+import android.os.Handler
+import android.os.Looper
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.Future
+
+/**
+ * Provides the entry point for, and management of, thumbnail extraction work.
+ */
+class VideoThumbnailExtractor @JvmOverloads constructor(
```
ReallyVasiliy, line 13: Add experimental annotation

litr/src/main/java/com/linkedin/android/litr/render/GlThumbnailRenderer.kt
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    fun saveTexture(width: Int, height: Int): Bitmap {
+        val capacity = width * height * 4
+        if (pixelBuffer == null || pixelBuffer!!.capacity() != capacity) {
+            pixelBuffer = ByteBuffer.allocate(capacity)
+        } else {
+            pixelBuffer?.rewind()
+        }
```
ReallyVasiliy, line 55: Make private?
Initialize buffer once at init, instead of lazily

litr/src/main/java/com/linkedin/android/litr/thumbnails/ThumbnailExtractParameters.kt
```diff
@@ -0,0 +1,50 @@
...
+        val renderer: ThumbnailRenderer,
+        /**
+         * A hint to the [ExtractionBehavior] about what approach to use for extracting video frames.
+         */
+        val mode: ExtractionMode = ExtractionMode.Fast,
+        /**
+         * Optional size of the generated thumbnail Bitmap. The video frame will be resized and scaled to fill this [destSize], preserving the video frame's
+         * original aspect ratio.
+         */
+        val destSize: Point? = null,
```
izzytwosheds, line 44: Since we are preserving original aspect ratio, passing a resolution (smaller dimension) should be better. What happens when `destSize`'s aspect ratio is different from video's? Or should we just scale the video to fit `destSize` like when we do in transcoding?

litr/src/main/java/com/linkedin/android/litr/thumbnails/ThumbnailExtractParameters.kt
```diff
@@ -0,0 +1,50 @@
...
+        /**
+         * The optional priority of the request. Lower value indicates higher priority. Requests with equal priority will be handled in FIFO order.
+         */
+        val priority: Long = 0L
```
izzytwosheds, line 48: Is there a range of these? Does this match thread executor priority api?

litr/src/main/java/com/linkedin/android/litr/thumbnails/VideoThumbnailExtractor.kt
```diff
@@ -0,0 +1,144 @@
...
+                    func(listener)
+                }
+            }
+        }
+    }
+
+    private data class ActiveExtractJob(val future: Future<*>, val listener: ThumbnailExtractListener?)
+
+    companion object {
+        private val TAG = VideoThumbnailExtractor::class.qualifiedName
```
izzytwosheds, line 142: Should this just be `private const val TAG = "VideoThumbnailExtractor"`?

litr/src/main/java/com/linkedin/android/litr/thumbnails/queue/PriorityThreadExecutor.kt
```diff
@@ -0,0 +1,20 @@
...
+package com.linkedin.android.litr.thumbnails.queue
+
+import java.util.concurrent.*
+
+/**
+ * Helpers for constructing an [ExecutorService] backed by a [PriorityBlockingQueue] for prioritizing tasks.
+ */
+object PriorityExecutorUtil {
+    fun newSingleThreadPoolPriorityExecutor() = ThreadPoolExecutor(
+        1, 1, 0L,
```
izzytwosheds, line 17: nip: these should probably be on separate lines, they seem to be all over the place.

litr/src/main/java/com/linkedin/android/litr/thumbnails/ThumbnailExtractParameters.kt
ReallyVasiliy: Let me add this image just so we're using the same language around this
image

> passing a resolution (smaller dimension) should be better

Can you clarify what you mean by resolution? Do you mean we would perform "aspect fit" scaling instead of "aspect fill"?

> What happens when `destSize`'s aspect ratio is different from video's?

Right now it will scale with "aspect fill": the video frame will be resized and cropped to _fill_ the `destSize` area, and the aspect ratio of the video frame is preserved.

litr/src/main/java/com/linkedin/android/litr/thumbnails/ThumbnailExtractParameters.kt
ReallyVasiliy: The range is Long.MIN_VALUE..Long.MAX_VALUE.

This value is used to set the priority of this request relative to other requests you've submitted, so it's completely arbitrary. We could add more structure around this by e.g. making it an enum, though that seemed a little prescriptive (not sure how consumers would choose to set priority). 

We use a `ThreadPoolExecutor` with a `PriorityBlockingQueue` work queue. We call `execute()` with an instance of our `ComparableFutureTask` (which overrides `comparedTo` to use the priority value), and this task gets added to the `PriorityBlockingQueue`. So when the executor polls the next task, it would always be the one with the lowest `priority` value.

litr/src/main/java/com/linkedin/android/litr/thumbnails/ThumbnailExtractParameters.kt
ReallyVasiliy: Actually, do you mean we should do this? -- 
1. the resulting bitmap will always have the same aspect ratio as the source video (i.e. we won't add any whitespace/padding)
2. instead of "destination size", we pass in something like "maximum width" and "maximum height"

izzytwosheds: Originally, I was thinking of doing this:
 - thumbnail bitmap will have the aspect ratio of video
 - by default, thumbnail will also have a resolution of source video (true frame preview)
 - optionally we can pass the "resolution" parameter, thumbnail bitmap's smaller dimension will be equal to resolution. This is closer to how video is defined on Android - resolution + rotation in 90 degree increments. 

For thumbnail use case, it is probably reasonable to aspect fill (AKA center crop) the thumbnail, as long as it is clearly defined in the API.

litr/src/main/java/com/linkedin/android/litr/thumbnails/ThumbnailExtractParameters.kt
izzytwosheds: Got it now, thanks!

litr/src/main/java/com/linkedin/android/litr/render/GlThumbnailRenderer.kt
```diff
@@ -0,0 +1,161 @@
...
+import com.linkedin.android.litr.filter.GlFrameRenderFilter
+import com.linkedin.android.litr.filter.video.gl.DefaultVideoFrameRenderFilter
+import java.nio.ByteBuffer
+import java.util.*
+
+
+/**
+ * A renderer that applies OpenGL filters to a bitmap, and returns a new bitmap.
+ */
+class GlThumbnailRenderer(filters: List<GlFilter>?) : ThumbnailRenderer {
```
izzytwosheds, line 27: Note: In the future we should refactor this to use `GlVideoRenderer`. Two renderers behave differently, what this renderer produces is not always a true representation of end video.

litr/src/main/java/com/linkedin/android/litr/thumbnails/VideoThumbnailExtractor.kt
```diff
@@ -0,0 +1,144 @@
...
+ * Provides the entry point for thumbnail extraction.
+ *
+ * This class uses a single, dedicated thread to schedule jobs. The priority of each job can be specified within [ThumbnailExtractParameters].
+ *
+ * @param context The application context.
+ * @param listenerLooper The looper on which [extract] listener events will be processed.
+ * @param extractionBehavior The behavior to use for extracting frames from media.
+ */
+@ExperimentalThumbnailsApi
+class VideoThumbnailExtractor @JvmOverloads constructor(
```
izzytwosheds, line 35: Can we rename this (and related stack) to `VideoFrameExtractor`? This should be a general purpose utility, thumbnails being one of use cases.

litr/src/main/java/com/linkedin/android/litr/render/GlThumbnailRenderer.kt
ReallyVasiliy: > Two renderers behave differently, what this renderer produces is not always a true representation of end video.

That's not ideal nor intended obviously. My rationale for having the separate renderer was based on the assumption that: a renderer is responsible only for providing the rendering resources (e.g. GL context, texture, vertex transformation, etc) needed for filters to draw to some target surface, and the actual rendering rules are defined in filters themselves. The two renderers didn't appear to have much in common in that respect. The main difference to consider was the way vertex transformation is defined -- but since the GlThumbnailRenderer always outputs the frame in the same aspect ratio, and because video is already rotated (AFAIK rotation is handled by the extractor for the respective media type), the matrix doesn't scale nor rotate

izzytwosheds: I see your reasoning. I still think we can try to consolidate this in the future, but for now this is ok.

litr/src/main/java/com/linkedin/android/litr/Annotations.kt
```diff
@@ -0,0 +1,9 @@
+package com.linkedin.android.litr
+
+@Retention(value = AnnotationRetention.BINARY)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING,
+    message = "The frame extract APIs are experimental in LiTr, and may be changed or removed in the future."
+)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.TYPEALIAS, AnnotationTarget.PROPERTY)
+annotation class ExperimentalFrameExtractorApi
```
izzytwosheds, line 9: This is very nifty way to mark experimental APIs. Let's generalize this, by renaming it to '@ExperimentalApi`

litr/src/main/java/com/linkedin/android/litr/frameextract/FrameExtractJob.kt
```diff
@@ -0,0 +1,110 @@
...
+import android.media.ThumbnailUtils
+import android.util.Log
+import com.linkedin.android.litr.ExperimentalFrameExtractorApi
+import com.linkedin.android.litr.frameextract.behaviors.FrameExtractBehavior
+import com.linkedin.android.litr.frameextract.behaviors.FrameExtractBehaviorFrameListener
+
+/**
+ * Provides the request lifecycle for extracting video frames. The specifics of extraction work are delegated to [FrameExtractBehavior]s.
+ */
+@ExperimentalFrameExtractorApi
```
izzytwosheds, line 20: This is exposed externally? We should make these `internal`
Note to self: In general, we should change our build config to enforce explicit access qualifiers in Kotlin classes, to prevent clients from using stuff that is not meant to be exposed.

litr/src/main/java/com/linkedin/android/litr/frameextract/behaviors/MediaMetadataExtractBehavior.kt
```diff
@@ -0,0 +1,70 @@
...
+package com.linkedin.android.litr.frameextract.behaviors
+
+import android.content.Context
+import android.media.MediaMetadataRetriever
+import android.net.Uri
+import com.linkedin.android.litr.ExperimentalFrameExtractorApi
+import com.linkedin.android.litr.frameextract.FrameExtractMode
+import com.linkedin.android.litr.frameextract.FrameExtractParameters
+
+@ExperimentalFrameExtractorApi
```
izzytwosheds, line 17: Do we still need this here? Does this get inherited from parent interface?

litr/src/main/java/com/linkedin/android/litr/render/SingleFrameRenderer.kt
```diff
@@ -0,0 +1,14 @@
...
+ * All Rights Reserved.
+ *
+ * Licensed under the BSD 2-Clause License (the "License").  See License in the project root for
+ * license information.
+ */
+package com.linkedin.android.litr.render
+
+import android.graphics.Bitmap
+
+interface SingleFrameRenderer {
```
izzytwosheds, line 12: Can we mark this as experimental?

litr/src/main/java/com/linkedin/android/litr/Annotations.kt
ReallyVasiliy: Rationale for making experimental annotation specific to a feature was that, if we add a new experimental feature, then consumers might unknowingly be opting into it (e.g. when someone opts into `@OptIn(LiTrExperimentalApi::class)`, and we add a new feature `FooExperiment`, they won't get any warnings)

litr/src/main/java/com/linkedin/android/litr/frameextract/behaviors/MediaMetadataExtractBehavior.kt
ReallyVasiliy: Yeah I believe we still need this -- basically this is propagating the opt-in requirement to whoever is using this concrete behavior. If I leave this out, we'll get warnings in this class that it's using an experimental API. So we have two choices: opt in, or propagate. If we opt in here, then callers won't know this class is using experimental APIs. So instead we propagate the requirement. https://kotlinlang.org/docs/opt-in-requirements.html#propagating-opt-in

litr/src/main/java/com/linkedin/android/litr/frameextract/FrameExtractJob.kt
ReallyVasiliy: Ah good catch, done. I also made the executor/task related classes internal (they were originally being injected so they were open)

izzytwosheds (approved):

PR #154, opened 2022-01-06
Fix for incorrect size estimation with range

izzytwosheds:
Duration estimation had a bug, it didn't work when media range started at non-zero and ended at Long.MAX_VALUE


ReallyVasiliy (approved):
This makes sense to me, I guess the only consideration since we're now clamping trim range to [0..duration], is track format must have KEY_DURATION. But I believe that'll always be the case right?

izzytwosheds:
It is possible not to have KEY_DURATION value, but that seems to be very rare. We should probably look into better way of handling this.

PR #159, opened 2022-01-18
New AudioRenderer implementation, with a render queue

izzytwosheds:
Basic implementation of audio renderer. Doesn't do much yet (no resampling, no mixing), but does couple of important things:
 - has a render queue which queues up audio frames in order, while encoder is busy and cannot accept new frames
 - handles the case when size of decoded buffer is larger than size of encoder's input buffer by sending decoded buffer into encoder in several parts

Not having render queue in existing audio renderer does result in dropped frames which manifests as distorted audio.

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/render/AudioRenderer.kt
+++ b/litr/src/main/java/com/linkedin/android/litr/render/AudioRenderer.kt
@@ -1,3 +1,10 @@
+/*
+ * Copyright 2019 LinkedIn Corporation
+ * All Rights Reserved.
+ *
+ * Licensed under the BSD 2-Clause License (the "License").  See License in the project root for
+ * license information.
+ */
 package com.linkedin.android.litr.render
 
 import android.media.MediaCodec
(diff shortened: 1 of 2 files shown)


litr/src/main/java/com/linkedin/android/litr/render/AudioRenderer.kt
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        private fun renderFrame(tag: Int, inputFrame: Frame) {
+            encoder.getInputFrame(tag)?.let { outputFrame ->
+                if (outputFrame.buffer != null && inputFrame.buffer != null) {
+                    outputFrame.bufferInfo.offset = 0
+                    outputFrame.bufferInfo.flags = inputFrame.bufferInfo.flags
+                    outputFrame.bufferInfo.presentationTimeUs =
+                        inputFrame.bufferInfo.presentationTimeUs +
+                            ((inputFrame.buffer.position() / (channelCount * BYTES_PER_SAMPLE)) * sampleDurationUs).toLong()
```
lopeng, line 109: BYTES_PER_SAMPLE could be 8, 16, 32, right?

izzytwosheds: No, on Android it is only 16. MediaCodec uses Short or recently 16 bit Float. There is support for 24 bit audio, but only though some audio specific APIs, not MediaCodec.

lopeng (approved):

PR #160, opened 2022-01-21
Audio resampling

izzytwosheds:
Using Oboe to resample audio track:
 - Oboe resampler (https://github.com/google/oboe/tree/master/src/flowgraph/resampler) is copied and used via JNI
 - `AudioRenderer` is modified to send a source buffer to Oboe, receive a resized buffer and queue it to be encoded.

### DIFF
--- a/litr/src/main/cpp/audio-resampler.cpp
+++ b/litr/src/main/cpp/audio-resampler.cpp
@@ -35,9 +35,10 @@ Java_com_linkedin_android_litr_render_AudioRenderer_resample(
         auto inputBuffer = new float[sampleCount * channel_count];
         auto outputBuffer = new float[channel_count];
 
+        // bytes contained in audio buffer produced by MediaCodec contains make up big endian shorts
+        // first we recreate short values, then simply cast them to floats, expected by Oboe resampler
         for (int index = 0; index < sampleCount * channel_count; index++) {
-            short value = (sourceBuffer[index * 2] << 8) | sourceBuffer[index * 2 + 1];
-            inputBuffer[index] = ((float) value);
+            inputBuffer[index] = (float) ((short) ((sourceBuffer[index * 2] << 8) | sourceBuffer[index * 2 + 1]));
         }
 
         int framesProcessed = 0;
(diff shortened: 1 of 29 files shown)


litr/src/main/cpp/audio-resampler.cpp
```diff
@@ -0,0 +1,77 @@
...
+        auto sourceBuffer = (jbyte *) env->GetDirectBufferAddress(jsourceBuffer);
+        jshort* targetBuffer = env->GetShortArrayElements(jtargetBuffer, JNI_FALSE);
+
+        auto inputBuffer = new float[sampleCount * channel_count];
+        auto outputBuffer = new float[channel_count];
+
+        // bytes contained in audio buffer produced by MediaCodec contains make up big endian shorts
+        // first we recreate short values, then simply cast them to floats, expected by Oboe resampler
+        for (int index = 0; index < sampleCount * channel_count; index++) {
+            inputBuffer[index] = (float) ((short) ((sourceBuffer[index * 2] << 8) | sourceBuffer[index * 2 + 1]));
```
philburk, line 41: jbytes are signed. So the bytes will get sign extended, which corrupts the higher bytes.
Either mask the signed bytes with 0xFF or make the sourceBuffer a (uint8_t *).

litr/src/main/cpp/audio-resampler.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+        while (inputFramesLeft > 0) {
+            if (audio_resampler->isWriteNeeded()) {
+                audio_resampler->writeNextFrame(inputBuffer);
+                inputBuffer += channel_count;
+                inputFramesLeft--;
+            } else {
+                audio_resampler->readNextFrame(outputBuffer);
+                for (int channel = 0; channel < channel_count; channel++) {
+                    targetBuffer[framesProcessed * channel_count + channel] = (short) *(outputBuffer + channel);
```
philburk, line 55: Resampling of a loud full amplitude signal can create signals out of range.
Those will wrap and sound bad.
So you should clip the float values between -32768 and 32767 before converting to float.

izzytwosheds:
This PR was broken into separate PRs. Closing it.

PR #163, opened 2022-01-26
Add resampling support in AudioRenderer

izzytwosheds:
With `OboeAudioResampler` implemented, we can now add audio resampling support in `AudioRenderer`:
 - current logic of copying source `Frame` into target one is abstracted out into `PassthroughResampler`
 - `AudioResamplerFactory` is added to create correct implementation of `AudioResampler` depending if resampling support is needed
 - `AudioResampler` is now used by default

### DIFF
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
@@ -23,7 +23,6 @@ import androidx.annotation.Nullable;
 
 import com.linkedin.android.litr.MediaTransformer;
 import com.linkedin.android.litr.TrackTransform;
-import com.linkedin.android.litr.TransformationListener;
 import com.linkedin.android.litr.TransformationOptions;
 import com.linkedin.android.litr.codec.MediaCodecDecoder;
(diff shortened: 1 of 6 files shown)


litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
```diff
@@ -568,7 +567,7 @@ public void transcodeAudio(@NonNull SourceMedia sourceMedia,
                 AudioTrackFormat trackFormat = (AudioTrackFormat) targetTrack.format;
                 mediaFormat = MediaFormat.createAudioFormat(
                         "audio/mp4a-latm",
-                        trackFormat.samplingRate,
+                        trackFormat.samplingRate / 2,
```
lopeng, line 570: I would prefer picking common samplingRates for demo, e.g., 44k -> 8k, or other pairs, instead of a simple half (48K -> 24K?), which might also help find out issue early).

izzytwosheds: The idea here was to guarantee resampling, regardless of source sample rate. I should probably make this configurable in UI.

PR #167, opened 2022-01-27
Rename AudioResampler to AudioProcessor

izzytwosheds:
`AudioResampler` is no longer just a resampler. It now also mixes channels and could more more in the future. So we are giving it a bit more "general purpose" name of `AudioProcessor`. The rest of the stack (native, etc.) is also renamed, methods are renamed and javadocs are updated.


litr/src/main/java/com/linkedin/android/litr/render/AudioProcessor.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2019 LinkedIn Corporation
```
lopeng, line 2: you need update the year in a PR later

lopeng (approved):

litr/src/main/java/com/linkedin/android/litr/render/AudioProcessor.kt
izzytwosheds: done

litr/src/main/java/com/linkedin/android/litr/render/OboeAudioProcessor.kt
```diff
@@ -17,35 +17,35 @@ import kotlin.math.ceil
 private const val BYTES_PER_SAMPLE = 2
 
 /**
- * Implementation of audio resampler that uses Oboe library
+ * Implementation of audio processor that uses Oboe library
  */
-internal class OboeAudioResampler(
+internal class OboeAudioProcessor(
```
ReallyVasiliy, line 22: Just feedback on naming

I think there are two paths/patterns in the mix here. 

One pattern is where an "audio processor" must be able to support all operations, like resample, mix, etc. In this case, maybe instead of `processFrame`, we would keep the function name at `resample`.

The other one, which is what it looks like you're doing, is where the "processor" is more of a single operation. So in this case, it could make sense to name this something like `OboeAudioResampler`.

izzytwosheds: So initially I was thinking of splitting processing, mixing, etc. into separate interfaces, but then realized that that would result in ton of extra byte buffer allocations and decided to have everything as "processing" in native layer. I think, realistically I will not be adding anything else to processing - resampling and muxing is all we need. All other transformations/effects will be implemented as filters, like we do with video frames.

PR #169, opened 2022-01-31
AudioRenderer unit tests

izzytwosheds:
Unit tests for `AudioRenderer`. We cannot test resampling in JVM tests because we use JNI for that. We will write a separate set of tests for that in instrumentation layer. 

We had to change how `BufferInfo` values are set in `PassthroughAudioProcessor` since `set` method doesn't seem to work in JVM tests.

### DIFF
--- a/litr/src/test/java/com/linkedin/android/litr/render/AudioRendererShould.kt
+++ b/litr/src/test/java/com/linkedin/android/litr/render/AudioRendererShould.kt
@@ -1,0 +1,187 @@
+package com.linkedin.android.litr.render
+
+import android.media.MediaCodec
+import android.media.MediaFormat
+import com.linkedin.android.litr.codec.Encoder
+import com.linkedin.android.litr.codec.Frame
+import org.hamcrest.CoreMatchers.equalTo
+import org.hamcrest.MatcherAssert.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.mockito.MockitoAnnotations
+import org.mockito.kotlin.any
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import java.nio.ByteBuffer
+import java.util.concurrent.CountDownLatch
+import kotlin.test.assertNotNull
+
+private const val INPUT_FRAME_TAG = 0
+private const val OUTPUT_FRAME_TAG_1 = 1
+private const val OUTPUT_FRAME_TAG_2 = 2
+private const val INPUT_BUFFER_SIZE = 42
+private const val PRESENTATION_TIME = 100L
+private const val ENCODER_BUFFER_CAPACITY = 42
+private const val EXTRA_INPUT_BYTES = 10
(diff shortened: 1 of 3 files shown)


litr/src/test/java/com/linkedin/android/litr/render/AudioRendererShould.kt
```diff
@@ -0,0 +1,187 @@
...
+    private val encoderInputFrame1 =
+        Frame(OUTPUT_FRAME_TAG_1, ByteBuffer.allocate(ENCODER_BUFFER_CAPACITY), MediaCodec.BufferInfo())
+
+    private val encoderInputFrame2 =
+        Frame(OUTPUT_FRAME_TAG_2, ByteBuffer.allocate(ENCODER_BUFFER_CAPACITY), MediaCodec.BufferInfo())
+
+    private lateinit var audioRenderer: AudioRenderer
+
+    private lateinit var encoderLatch: CountDownLatch
+    private val outputFrames = mutableListOf<Frame>()
```
ReallyVasiliy, line 44: nit: you can potentially avoid having this state up here by moving this to a method that returns something like a future/callable on completion, e.g.
```kotlin
fun Encoder.collectOutputFrames(count: Int): Callable<List<Frame>> {
    val latch = CountDownLatch(count)
    val frames = mutableListOf<Frame>()
    whenever(this.queueInputFrame(any())).thenAnswer {
        frames.add(it.arguments[0] as Frame)
        latch.countDown()
    }
    return Callable {
        latch.await()
        frames
    }
}
```

Then you would do something like,
```kotlin
val collect = encoder.collectOutputFrames(2)
// ... act ...
val outputFrames = collect.call()
```

ReallyVasiliy (approved):
A few minor comments, but LGTM! 👍

litr/src/test/java/com/linkedin/android/litr/render/AudioRendererShould.kt
```diff
@@ -0,0 +1,187 @@
...
+
+        // first encoder input frame should be filled and start at input presentation time
+        val expectedByteBuffer1 = ByteBuffer.allocate(ENCODER_BUFFER_CAPACITY)
+        repeat(ENCODER_BUFFER_CAPACITY) {
+            expectedByteBuffer1.put(it.toByte())
+        }
+        val expectedBufferInfo1 = MediaCodec.BufferInfo()
+        expectedBufferInfo1.offset = 0
+        expectedBufferInfo1.size = ENCODER_BUFFER_CAPACITY
+        expectedBufferInfo1.presentationTimeUs = inputFrame.bufferInfo.presentationTimeUs
```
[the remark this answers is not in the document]
izzytwosheds: Good call, done.

litr/src/test/java/com/linkedin/android/litr/render/AudioRendererShould.kt
izzytwosheds: Wow, this is a lot more elegant than what I have. Will change my code to use this.

PR #170, opened 2022-02-03
Passthrough ByteBuffer encoder

izzytwosheds:
An implementation of `Encoder` which simply passes input `ByteBuffer`s into output, with no modification. Useful when raw decoded frames are needed - for inspection, or to write audio frames into a WAV file. 

Demo of writing audio into WAV file is included.

### DIFF
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
@@ -24,9 +24,11 @@ import androidx.annotation.Nullable;
 import com.linkedin.android.litr.MediaTransformer;
 import com.linkedin.android.litr.TrackTransform;
 import com.linkedin.android.litr.TransformationOptions;
+import com.linkedin.android.litr.codec.Encoder;
 import com.linkedin.android.litr.codec.MediaCodecDecoder;
 import com.linkedin.android.litr.codec.MediaCodecEncoder;
 import com.linkedin.android.litr.codec.PassthroughDecoder;
+import com.linkedin.android.litr.codec.PassthroughBufferEncoder;
 import com.linkedin.android.litr.exception.MediaTransformationException;
 import com.linkedin.android.litr.filter.GlFilter;
 import com.linkedin.android.litr.filter.GlFrameRenderFilter;
@@ -38,6 +40,8 @@ import com.linkedin.android.litr.io.MediaRange;
 import com.linkedin.android.litr.io.MediaSource;
 import com.linkedin.android.litr.io.MediaTarget;
 import com.linkedin.android.litr.io.MockVideoMediaSource;
+import com.linkedin.android.litr.io.WavMediaTarget;
+import com.linkedin.android.litr.render.AudioRenderer;
 import com.linkedin.android.litr.render.GlVideoRenderer;
 import com.linkedin.android.litr.utils.TransformationUtil;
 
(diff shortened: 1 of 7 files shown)


litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
```diff
@@ -556,33 +560,46 @@ public void transcodeAudio(@NonNull SourceMedia sourceMedia,
...
                 : new MediaRange(0, Long.MAX_VALUE);
 
-        TransformationOptions transformationOptions = new TransformationOptions.Builder()
-                .setGranularity(MediaTransformer.GRANULARITY_DEFAULT)
-                .setSourceMediaRange(mediaRange)
-                .build();
+        try {
+            String targetMimeType = targetMedia.writeToWav ? "audio/raw" : "audio/mp4a-latm";
+            MediaTarget mediaTarget = targetMedia.writeToWav
+                    ? new WavMediaTarget(targetMedia.targetFile.getPath())
```
ReallyVasiliy, line 566: minor: Should we change the extension to wav in these cases?

litr/src/main/java/com/linkedin/android/litr/codec/PassthroughBufferEncoder.kt
```diff
@@ -0,0 +1,128 @@
...
+            encodedFrames[frame.tag] = frame
+        }
+    }
+
+    override fun dequeueOutputFrame(timeout: Long): Int {
+        if (!mediaFormatChanged) {
+            mediaFormatChanged = true
+            return MediaCodec.INFO_OUTPUT_FORMAT_CHANGED
+        }
+        return encodeQueue.poll() ?: MediaCodec.INFO_TRY_AGAIN_LATER
```
ReallyVasiliy, line 96: Should we be passing the timeout to poll() here? I realize this is passthrough but are we sure no one's going to have a race condition queuing/dequeuing input/output frames?

litr/src/main/java/com/linkedin/android/litr/codec/PassthroughBufferEncoder.kt
```diff
@@ -0,0 +1,128 @@
...
+        return dequeuedInputFrames[tag]
+    }
+
+    override fun queueInputFrame(frame: Frame) {
+        // since this method modifies multiple data structures, we have to make it synchronous
+        synchronized(this) {
+            frame.buffer?.flip()
+            dequeuedInputFrames.remove(frame.tag)
+            encodeQueue.add(frame.tag)
+            encodedFrames[frame.tag] = frame
```
ReallyVasiliy, line 87: nit: should the order of these two lines be reversed? (queue the frame after it can be read)

litr/src/main/java/com/linkedin/android/litr/codec/PassthroughBufferEncoder.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    override fun getOutputFrame(tag: Int): Frame? {
+        return encodedFrames[tag]
+    }
+
+    override fun releaseOutputFrame(tag: Int) {
+        encodedFrames.remove(tag)?.let {
+            it.buffer?.clear()
+            availableFrames.add(it)
+        }
```
ReallyVasiliy, line 107: If we're using synchronize for transactions, should we do that here too?

litr/src/main/java/com/linkedin/android/litr/codec/PassthroughBufferEncoder.kt
ReallyVasiliy: To clarify the previous comment a bit -- 
Because `timeout` is specified in the API, it's possible that someone might rely on it, so I'm wondering if it's something we should support even in this passthrough encoder.

litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
izzytwosheds: I tried doing that, but for some reason transcoded file becomes inaccessible.

litr/src/main/java/com/linkedin/android/litr/codec/PassthroughBufferEncoder.kt
izzytwosheds: Good call. Added.

PR #171, opened 2022-02-03
Capability to apply filters in AudioRenderer

izzytwosheds:
Similar to video GL filters, adding support for audio filters:
 - introducing a new `BufferFilter` interface that is called to modify `Frame` contents, meant to be used for any `Renderer` in a buffer mode
 - `AudioRenderer` can now accept a list of filters and apply them in order.
 - No actual filters are yet implemented

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/filter/BufferFilter.kt
+++ b/litr/src/main/java/com/linkedin/android/litr/filter/BufferFilter.kt
@@ -1,0 +1,17 @@
+package com.linkedin.android.litr.filter
+
+import com.linkedin.android.litr.codec.Frame
+
+/**
+ * A filter used with a renderer operating in buffer mode, like [AudioRenderer]
+ */
+interface BufferFilter {
+
+    /**
+     * Apply a filter to a [Frame]. Frame.bufferInfo will provide necessary metadata.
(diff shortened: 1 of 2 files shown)


litr/src/main/java/com/linkedin/android/litr/filter/BufferFilter.kt
```diff
@@ -0,0 +1,17 @@
...
+ */
+interface BufferFilter {
+
+    /**
+     * Apply a filter to a [Frame]. Frame.bufferInfo will provide necessary metadata.
+     * Frame.buffer is expected to be modified. Buffer contents are in target format.
+     * For example, for audio buffer those will be Short (2 byte LE) values with target
+     * sample rate and channel count.
+     */
+    fun apply(frame: Frame)
```
ReallyVasiliy, line 16: LGTM, but I wonder if this will need some details about the target format (e.g. sample rate, channel count) for concrete implementations to build useful filters

ReallyVasiliy (approved):

litr/src/main/java/com/linkedin/android/litr/filter/BufferFilter.kt
izzytwosheds: Good call, I was just thinking about that. Added.

PR #172, opened 2022-02-03
Pass audio filters in TransformationOptions

izzytwosheds:
We only supported passing in audio filters in `TrackTransform` api. We now support it in simpler api as well.

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/render/AudioRenderer.kt
+++ b/litr/src/main/java/com/linkedin/android/litr/render/AudioRenderer.kt
@@ -26,7 +26,7 @@ private const val TAG = "AudioRenderer"
 
 class AudioRenderer @JvmOverloads constructor(
     private val encoder: Encoder,
-    filters: MutableList<BufferFilter>? = mutableListOf()
+    filters: MutableList<BufferFilter>? = null
 ) : Renderer {
 
     private val filters: List<BufferFilter> = filters ?: listOf()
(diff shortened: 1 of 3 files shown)


litr/src/main/java/com/linkedin/android/litr/render/AudioRenderer.kt
```diff
@@ -26,9 +26,11 @@ private const val TAG = "AudioRenderer"
 
 class AudioRenderer @JvmOverloads constructor(
     private val encoder: Encoder,
-    private val filters: MutableList<BufferFilter> = mutableListOf()
+    filters: MutableList<BufferFilter>? = null
```
ReallyVasiliy, line 29: Should this be `private val filters: List<BufferFilter> = listOf()`?

We're requiring that this list is mutable, but we're always just casting it as an immutable list.

izzytwosheds: So we want the list to be immutable internally, but cannot accept an immutable list in constructor, because Java doesn't like it. Therefore, all this finagling.

ReallyVasiliy: OK 👍  but would be curious to see examples where this causes issues in Java if you come across them in the future, as I haven't seen issues around this before

ReallyVasiliy (approved):

PR #177, opened 2022-03-04
Ensure non-started jobs are removed from jobs map when stopping

svanegas:
### The problem

Whenever a frame extraction gets cancelled by its request id, there were some cases when a subsequent extract call for the same request id was not reporting any result/error/cancellation.

### The cause

The initial cancellation call was fired when the extraction job hadn't started yet, this was causing the job never being interrupted and never getting into a point of triggering a cancellation listener. The cancellation listener is the one in charge of removing the job from the `activeJobMap`, so since it was never removed, subsequent calls with the same request id were immediately discarded because the job still was in the `activeJobMap`.

### The solution

Add a `isStarted` property to the `FrameExtractJob`, which gets set to `true` right at the beginning of the `run` method of the `Runnable`. This property is exposed to `ComparableFutureTask`, which at the same time can be read from the `ActiveExtractJob` in order to determine when the job is being cancelled but hasn't started yet.

### Visuals

Here's an example of this behavior, modifying a bit the `litr-demo` we can observe the issue and after the fix.
- Caching was removed for better visualization
- A request id is not random but is uniquely mapped to each frame (based on the timestamp)

| Before | After |
|----|----|
| before | after |

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/frameextract/FrameExtractJob.kt
+++ b/litr/src/main/java/com/linkedin/android/litr/frameextract/FrameExtractJob.kt
@@ -24,6 +24,10 @@ internal class FrameExtractJob constructor(
     private val behavior: FrameExtractBehavior,
     private val listener: FrameExtractListener?
 ) : Runnable {
+
+    var isStarted: Boolean = false
+        private set
+
     private val behaviorFrameListener = object: FrameExtractBehaviorFrameListener {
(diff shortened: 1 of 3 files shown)


ReallyVasiliy (approved):

izzytwosheds (approved):
Note to self: We should probably do something similar for `TransformationJob`s as well.

PR #185, opened 2022-04-12
Fixed angle computation

kolesnikov-pasha:
Just a wrong degree to radian formula


izzytwosheds:
Did this code actually run?

kolesnikov-pasha:
> Did this code actually run?

It shouldn't as I understand, but it might if user is using some not accurate math equations to compute angle. For example, it might be called if previously user computed angle from radians to degrees.

kolesnikov-pasha:
Or if user just want to rotate video in a strange way:)

izzytwosheds:
That angle is supposed to be video orientation (portrait, landscape, etc.), which is expected to be multiples of 90 degrees. But I guess one can expect any kind of user input. :)

PR #193, opened 2022-05-19
Audio filter that changes track volume

izzytwosheds:
Our first audio filter. Simplest one, changes track volume:
- takes a non-negative double value
- 0 volume silences the track
- volume 1 doesn't really change anything
- value <1 reduces the volume
- value >1 increases the volume
- actual sample value change is exponential, since volume perception is logarithmic

### DIFF
--- a/litr-filters/src/main/java/com/linkedin/android/litr/filter/audio/VolumeFilter.kt
+++ b/litr-filters/src/main/java/com/linkedin/android/litr/filter/audio/VolumeFilter.kt
@@ -21,7 +21,7 @@ private const val BASE = 10.0 // volume is logarithmic, we will use base 10
  * - 1 will keep the volume "as is"
  * - 0 will silence the track
  * - value <1 will lower the volume
- * - value >1 will increase it.
+ * - value >1 will increase it. One has to be careful with these, since large values may result in distortion.
  */
 class VolumeFilter(@FloatRange(from = 0.0) private val volume: Double) : BufferFilter{
 
--- a/litr-filters/src/main/java/com/linkedin/android/litr/filter/audio/VolumeFilter.kt
+++ b/litr-filters/src/main/java/com/linkedin/android/litr/filter/audio/VolumeFilter.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2022 LinkedIn Corporation
+ * All Rights Reserved.
+ *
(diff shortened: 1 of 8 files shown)


litr-filters/src/main/java/com/linkedin/android/litr/filter/audio/VolumeFilter.kt
```diff
@@ -0,0 +1,42 @@
...
+    override fun init(mediaFormat: MediaFormat?) {}
+
+    override fun apply(frame: Frame) {
+        frame.buffer?.asShortBuffer()?.let { buffer ->
+            val sampleCount = frame.bufferInfo.size / SAMPLE_SIZE
+            repeat(sampleCount) { index ->
+                // replace sample at index with volume adjusted value
+                buffer.put(
+                    index,
+                    (buffer.get(index) * (BASE.pow(volume) -1 ) / (BASE - 1)).toInt().toShort()
```
lopeng, line 37: Is this based on a common gain calculation? Hope it won't create unexpected noise/distortion.

litr-filters/src/main/java/com/linkedin/android/litr/filter/audio/VolumeFilter.kt
```diff
@@ -0,0 +1,42 @@
...
+private const val BASE = 10.0 // volume is logarithmic, we will use base 10
+
+/**
+ * An audio filter that changes the audio track volume:
+ * - 1 will keep the volume "as is"
+ * - 0 will silence the track
+ * - value <1 will lower the volume
+ * - value >1 will increase it. One has to be careful with these, since large values may result in distortion.
+ */
+class VolumeFilter(@FloatRange(from = 0.0) private val volume: Double) : BufferFilter{
```
lopeng, line 26: `Filter` doesn't seem associated with audio processing (thinking about sound effect, boost, mixer, etc). Possible better naming?

izzytwosheds: I couldn't really come up with better common naming for audio and video frame modifications. I guess "filter" works better with video because it became a common term with video/photo filters in popular apps. With audio "filters" have more of a "signal processing" meaning low/high pass filter, etc. Maybe we can keep "filter" terminology within LiTr?

lopeng (approved):

Issue #201, opened 2022-07-01
Maybe a memory leak in audio-process.cpp


kaitian521:
`
auto resamplerInputBuffer = new float[outputChannelCount];
auto resamplerOutputBuffer = new float[outputChannelCount];
`
This two float buffer can be allocated just by once， for example,  can be inited in `Java_com_linkedin_android_litr_render_OboeAudioProcessor_initProcessor`
and uninit in 
`Java_com_linkedin_android_litr_render_OboeAudioProcessor_releaseProcessor`
to avoid multiple memory junks

izzytwosheds:
Good call, I will look into this. Thank you!

PR #202, opened 2022-07-03
Allocate/deallocate native input/output buffers on init/release

izzytwosheds:
We were allocating native buffers for Oboe audio processor on each frame and not de-allocating them, as was pointed out in https://github.com/linkedin/LiTr/issues/201 
Instead, now we are allocating them when we initialize the resampler (when we know the buffer size) and deallocating them when we release the resampler.

### DIFF
--- a/litr/src/main/cpp/audio-processor.cpp
+++ b/litr/src/main/cpp/audio-processor.cpp
@@ -14,6 +14,9 @@ MultiChannelResampler* oboeResampler = nullptr;
 int inputChannelCount = -1;
 int outputChannelCount = -1;
 
+float* resamplerInputBuffer = nullptr;
+float* resamplerOutputBuffer = nullptr;
+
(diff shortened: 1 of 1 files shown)


litr/src/main/cpp/audio-processor.cpp
```diff
@@ -94,6 +97,14 @@ Java_com_linkedin_android_litr_render_OboeAudioProcessor_releaseProcessor(
         inputChannelCount = -1;
         outputChannelCount = -1;
     }
+    if (resamplerInputBuffer != nullptr) {
+        delete resamplerInputBuffer;
+        resamplerInputBuffer = nullptr;
+    }
+    if (resamplerOutputBuffer != nullptr) {
+        delete resamplerOutputBuffer;
```
ReallyVasiliy, line 105: nit: should these be `delete[]` (e.g. `delete[] resamplerInputBuffer`) instead?

ReallyVasiliy (approved):

PR #229, opened 2022-12-20
Implement AudioRecord support for audio track

IanDBird:
I'm hoping to introduce Camera support as an input into Litr. This would allow scenarios where the raw audio/video is not sourced from a static file, but instead coming from a "live" source. I am planning the following, separate, pull requests:
 - Audio Record Support: Ability to pull raw PCM from an audio source (e.g. microphone) to mux with video from another source
 - External Surface Support: Ability to expose the raw surface that is used by the filter pipeline, allowing some other component (e.g. Camera2, CameraX, 3rd party library) to render directly into this. This will essentially bypass the Demux/Decode steps of the pipeline.
 - Camera2/CameraX Support: Helper classes/source for most common camera source scenarios.

This first PR covers Audio Record Support. It introduces a new `AudioRecordMediaSource` class which is responsible for taking raw PCM from `AudioRecord`, computing the appropriate PTS and feeding into the rest of the pipeline via the `MediaSource` interface. For now, the Demo app is simply muxing this together with a blank/red video stream. In future updates, this will be combined with a Camera2 integration that will merge recorded video and audio into the same file.

### DIFF
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordAudioFragment.kt
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordAudioFragment.kt
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2022 LinkedIn Corporation
+ * All Rights Reserved.
+ *
+ * Licensed under the BSD 2-Clause License (the "License").  See License in the project root for
+ * license information.
+ *
+ * Author: Ian Bird
+ */
+package com.linkedin.android.litr.demo
+
+import android.Manifest
+import android.app.Activity
+import android.app.AlertDialog
+import android.content.pm.PackageManager
+import android.os.Build
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.RequiresApi
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
+import com.linkedin.android.litr.MediaTransformer
+import com.linkedin.android.litr.demo.data.TargetMedia
+import com.linkedin.android.litr.demo.data.TransformationPresenter
+import com.linkedin.android.litr.demo.data.TransformationState
+import com.linkedin.android.litr.demo.databinding.FragmentAudioRecordBinding
+import com.linkedin.android.litr.io.AudioRecordMediaSource
+import com.linkedin.android.litr.utils.TransformationUtil
+import java.io.File
+
+private const val REQUEST_AUDIO_RECORD_PERMISSION = 14
+
+class RecordAudioFragment : BaseTransformationFragment() {
+    private lateinit var binding: FragmentAudioRecordBinding
+
+    private lateinit var mediaTransformer: MediaTransformer
+    private var targetMedia: TargetMedia = TargetMedia()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        mediaTransformer = MediaTransformer(context!!.applicationContext)
+
+        // This demo fragment requires Android M or newer, in order to support reading data from
+        // AudioRecord in a non-blocking way. Let's double check that the current device supports
+        // this.
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            AlertDialog.Builder(requireContext())
+                    .setMessage("Android Marshmallow or newer required")
+                    .setPositiveButton("Ok") { _, _ ->
+                        activity?.onBackPressed()
+                    }
+                    .show()
+            return
+        }
+
+        // Check to make sure the user has granted permission to record audio.
+        if (!hasAudioRecordPermission()) {
+            ActivityCompat.requestPermissions(
+                    context as Activity,
+                    arrayOf(Manifest.permission.RECORD_AUDIO),
+                    REQUEST_AUDIO_RECORD_PERMISSION
+            )
+        }
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        mediaTransformer.release()
+    }
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    override fun onCreateView(
+            inflater: LayoutInflater,
+            container: ViewGroup?,
+            savedInstanceState: Bundle?
+    ): View {
+        binding = FragmentAudioRecordBinding.inflate(layoutInflater, container, false)
+
+        binding.transformationState = TransformationState()
+        binding.transformationPresenter = TransformationPresenter(context!!, mediaTransformer)
+        binding.mediaSource = AudioRecordMediaSource()
+
+        val targetFile = File(
+                TransformationUtil.getTargetFileDirectory(requireContext().applicationContext),
+                "recorded_audio_${System.currentTimeMillis()}.mp4"
+        )
+        targetMedia.setTargetFile(targetFile)
+        binding.targetMedia = targetMedia
+
+        return binding.root
+    }
+
+    private fun hasAudioRecordPermission(): Boolean {
+        val validContext = context ?: return false
+        return ContextCompat.checkSelfPermission(
+                validContext,
+                Manifest.permission.RECORD_AUDIO
+        ) == PackageManager.PERMISSION_GRANTED
+    }
+}
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
@@ -11,6 +11,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
+import android.graphics.Color;
 import android.graphics.PointF;
 import android.media.MediaFormat;
 import android.media.MediaMuxer;
@@ -39,6 +40,8 @@ import com.linkedin.android.litr.filter.GlFrameRenderFilter;
 import com.linkedin.android.litr.filter.Transform;
 import com.linkedin.android.litr.filter.audio.AudioOverlayFilter;
 import com.linkedin.android.litr.filter.video.gl.DefaultVideoFrameRenderFilter;
+import com.linkedin.android.litr.filter.video.gl.SolidBackgroundColorFilter;
+import com.linkedin.android.litr.io.AudioRecordMediaSource;
 import com.linkedin.android.litr.io.MediaExtractorMediaSource;
 import com.linkedin.android.litr.io.MediaMuxerMediaTarget;
 import com.linkedin.android.litr.io.MediaRange;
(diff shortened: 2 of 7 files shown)


litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordAudioFragment.kt
```diff
@@ -0,0 +1,93 @@
...
+    private var targetMedia: TargetMedia = TargetMedia()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        mediaTransformer = MediaTransformer(context!!.applicationContext)
+
+        // This demo fragment requires Android M or newer, in order to support reading data from
+        // AudioRecord in a non-blocking way. Let's double check that the current device supports
+        // this.
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
```
IanDBird, line 36: This requirement could be potentially removed in the future, but would require additional changes in `AudioRecordMediaSource`. We would need something like a ring buffer, with a separate thread reading via the older/blocking api and updating the buffer. Then we would read from this buffer in a non-blocking way. The downside is obviously additional complexity along with increased memory usage. For now, I thought requiring Android M or newer was better, at least for now.

litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
```diff
@@ -671,6 +674,82 @@ public void transcodeToVp9(@NonNull SourceMedia sourceMedia,
...
+        try {
+            MediaTarget mediaTarget = new MediaMuxerMediaTarget(targetMedia.targetFile.getPath(),
+                    2,
+                    0,
+                    MediaMuxer.OutputFormat.MUXER_OUTPUT_MPEG_4);
+
+            // Create a single (synthetic) video track to ensure our output is playable by the demo
+            // app. We only use 1 second of video (a solid color) so the duration of the video will
+            // likely depend on the length of the audio recording.
+            VideoTrackFormat videoTrackFormat = new VideoTrackFormat(0, MimeType.VIDEO_AVC);
```
IanDBird, line 699: To play a video-less video in the Demo app required additional changes. Therefore, I went with a simple approach of adding a short/synthetic video stream.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,179 @@
...
+        // If the recording has been stopped, then there are no more samples to be read.
+        if (!isRecording) {
+            return -1
+        }
+
+        // Read the next audio sample from the recorder. As per the AudioRecord documentation, we
+        // need to leave some buffer for the AudioRecord to continue to queue too. We also don't
+        // want to block waiting for more data, we'll take whatever is available.
+        val targetSize = (bufferSize / 2).coerceAtMost(buffer.capacity())
+        val readBytes = audioRecord?.read(buffer, targetSize, AudioRecord.READ_NON_BLOCKING) ?: -1
```
izzytwosheds, line 116: To correctly satisfy `MediaSource` contract we should use `offset` value here and read into `buffer` at that value: https://developer.android.com/reference/android/media/MediaExtractor#readSampleData(java.nio.ByteBuffer,%20int) 
This is working because `offset` is 0, but it might not always be. This can be in a follow-up PR.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,179 @@
...
+
+        totalDurationUs = 0L
+    }
+
+    override fun getSize() = -1L
+
+    companion object {
+        private val TAG = AudioRecordMediaSource::class.simpleName
+
+        private const val MEDIA_FORMAT_PCM_KEY = "pcm-encoding"
```
izzytwosheds, line 170: private `val`s don't really have to be inside companion object

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,179 @@
...
+        private const val AUDIO_FORMAT = AudioFormat.ENCODING_PCM_16BIT
+        private const val BYTES_PER_SAMPLE = 2
+
+        const val DEFAULT_AUDIO_SOURCE = MediaRecorder.AudioSource.MIC
```
izzytwosheds, line 174: can you make these `public` explicitly? They usually should be for libraries. (I should add that check)

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
izzytwosheds: Also, how is using `AudioRecord.READ_NON_BLOCKING` working? `readSampleData` is supposed to be a blocking call. 😅

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,179 @@
...
+ * from a given audio source (i.e. the device's microphone).
+ *
+ * This class requires Android M+ in order to support reading data from {@link AudioRecord} in a
+ * non-blocking way.
+ */
+@RequiresApi(Build.VERSION_CODES.M)
+class AudioRecordMediaSource(
+        private val audioSource: Int = DEFAULT_AUDIO_SOURCE,
+        private val sampleRate: Int = DEFAULT_SAMPLE_RATE,
+        private val channelConfig: Int = DEFAULT_CHANNEL_CONFIG,
```
izzytwosheds, line 22: It feels like it is possible to easily mismatch `channelConfig` and `channelCount`, which would cause problems, I presume. Should we just accept `channelConfig` and derive `channelCount` from it?

litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
izzytwosheds: That is totally fine.

litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordAudioFragment.kt
izzytwosheds: No need for anything that complicated, just leave things as is. Demo app is supposed to be a space for easy experimentation and sample code, not a prod ready app. What you are doing here is exactly what it is for.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
IanDBird: Will do. Have checked a few other files and noted that you move them before the class definition (e.g. `AudioOverlayFilter`). Will copy

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
IanDBird: I'm happy to add `public` but Android Studio will complain (since that's what it is by default). I don't see any other `public const val` in the code base, but no objections to adding if preferred.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
IanDBird: Something like this:

```
class AudioRecordMediaSource(
        private val audioSource: Int = DEFAULT_AUDIO_SOURCE,
        private val sampleRate: Int = DEFAULT_SAMPLE_RATE,
        private val isStereo: Boolean = false
) : MediaSource {

    private val channelCount: Int by lazy { 
        return@lazy if (isStereo) 2 else 1
    }
    
    private val channelConfig: Int by lazy {
        return@lazy if (isStereo) AudioFormat.CHANNEL_IN_STEREO else AudioFormat.CHANNEL_IN_MONO
    }
```

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
izzytwosheds: ok, then you can keep it as is. I will add library style enforcement later, perhaps.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
izzytwosheds: In that case I would probably keep the parameter an integer `channelCount` - masks internal `AudioConfig` implementation and is more future-proof (who knows, maybe someday Android devices will record a 6 channel sound?)

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
izzytwosheds: Yup. Kotlin doesn't really like static/companion things. Better use `const val` in class.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
izzytwosheds: Thank you for explanation, now this makes sense. So basically we are telling `AudioRecord` not to wait until it fills its internal buffer and fill ours with whatever it has at the moment, synchronously. This makes a lot of sense and should work - as long as size is not -1 `AudioTrackTranscoder` should keep going.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,207 @@
...
+ * This class requires Android M+ in order to support reading data from {@link AudioRecord} in a
+ * non-blocking way.
+ */
+@RequiresApi(Build.VERSION_CODES.M)
+class AudioRecordMediaSource(
+        private val audioSource: Int = DEFAULT_AUDIO_SOURCE,
+        private val sampleRate: Int = DEFAULT_SAMPLE_RATE,
+        private val isStereo: Boolean = false
+) : MediaSource {
+    private val channelCount: Int by lazy {
```
izzytwosheds, line 30: Not sure how much are we saving by using `lazy` here. There is no complex class initialization going on. I think we can remove all `lazy` intiatizations here and below.

IanDBird: Fair enough, I think I just got carried away :)

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
IanDBird: Yup, exactly.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
izzytwosheds: Been there, done that... many times... :)

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
IanDBird: Sure, will tweak 👍

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,220 @@
...
+) : MediaSource {
+
+    init {
+        // Check that we've been configured with a supported number of channels.
+        if (channelCount !in 1..2) {
+            throw IllegalArgumentException("Unsupported channel count configuration")
+        }
+    }
+
+    private val channelConfig =
```
izzytwosheds, line 47: we could probably do something like this here:
private val channelConfig = when (channelCount) {
   1 -> AudioFormat.CHANNEL_IN_MONO
   2 -> AudioFormat.CHANNEL_IN_STEREO
   else -> throw IllegalArgumentException("Unsupported channel count configuration")
}

This would get rid of init and make `AudioFormat` selection more explicit and readable.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,220 @@
...
+            if (channelCount == 1) AudioFormat.CHANNEL_IN_MONO else AudioFormat.CHANNEL_IN_STEREO
+
+    // Compute the appropriate buffer size based upon the audio configuration.
+    private val bufferSize: Int by lazy {
```
izzytwosheds, line 51: No need for `lazy`

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,220 @@
...
+        AudioRecord.getMinBufferSize(sampleRate, channelConfig, AUDIO_FORMAT) * 8
+    }
+
+    // Compute the number of bytes per microsecond of audio.
+    private val bytesPerUs: Double by lazy {
```
izzytwosheds, line 56: no need for `lazy`

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -0,0 +1,220 @@
...
+            position(original)
+        }
+    }
+
+    companion object {
+        const val DEFAULT_AUDIO_SOURCE = MediaRecorder.AudioSource.MIC
+        const val DEFAULT_SAMPLE_RATE = 44100
+        const val DEFAULT_CHANNEL_COUNT = 1
+    }
+}
```
izzytwosheds, line 220: nip: add empty last line

izzytwosheds (approved):

PR #231, opened 2022-12-26
Implement external surface support

IanDBird:
This PR introduces a new `ExternalMediaSource` class. This provides the owner direct access to the Surface that will be used as the input into any configured filters, followed by the encoder. The reason for this is to allow non-file based sources of video. In this example, i'm using Camera2 inside the DemoApp to render directly too it. This is being combined with the previously landed `AudioRecordMediaSource` that is adding the audio stream. This demonstrates a Camera to File example.

The DemoApp is currently doing a lot of Camera2 based configuration for this example. My plan/hope is to follow this up with another PR that introduces a `Camera2MediaSource` (and potentially a `CameraXMediaSource`). This will migrate some of the logic from the DemoApp into dedicated MediaSources, reducing the amount of code people need to write to utilise.

Also, one thing to note is that the `ExternalMediaSource` does not support cropping. It expects the source aspect ratio to match that of the file/target aspect ratio. We could either keep this as a hard constraint, or we could introduce a `CroppingFilter` to handle this case.

### DIFF
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCameraFragment.kt
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCameraFragment.kt
@@ -1,0 +1,358 @@
+/*
+ * Copyright 2022 LinkedIn Corporation
+ * All Rights Reserved.
+ *
+ * Licensed under the BSD 2-Clause License (the "License").  See License in the project root for
+ * license information.
+ *
+ * Author: Ian Bird
+ */
+package com.linkedin.android.litr.demo
+
+import android.Manifest
+import android.app.Activity
+import android.app.AlertDialog
+import android.content.Context
+import android.content.pm.PackageManager
+import android.content.res.Configuration
+import android.graphics.Rect
+import android.hardware.camera2.*
+import android.hardware.camera2.CameraCaptureSession.CaptureCallback
+import android.os.Build
+import android.os.Bundle
+import android.os.Handler
+import android.os.HandlerThread
+import android.util.Log
+import android.util.Range
+import android.util.Size
+import android.view.LayoutInflater
+import android.view.Surface
+import android.view.SurfaceHolder
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.RequiresApi
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
+import com.linkedin.android.litr.MediaTransformer
+import com.linkedin.android.litr.demo.data.TargetMedia
+import com.linkedin.android.litr.demo.data.TransformationPresenter
+import com.linkedin.android.litr.demo.data.TransformationState
+import com.linkedin.android.litr.demo.databinding.FragmentCameraRecordBinding
+import com.linkedin.android.litr.io.AudioRecordMediaSource
+import com.linkedin.android.litr.io.ExternalMediaSource
+import com.linkedin.android.litr.utils.TransformationUtil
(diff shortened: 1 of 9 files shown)


litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCameraFragment.kt
```diff
@@ -0,0 +1,358 @@
...
+private const val TAG = "RecordCameraFragment"
+
+private const val REQUEST_AUDIO_AND_CAMERA_PERMISSION = 27
+
+private const val DEFAULT_CAMERA_FPS = 30
+private const val DEFAULT_TARGET_BITRATE = 5_000_000 // 5Mbps
+private const val DEFAULT_RECORD_WIDTH = 1280
+
+@RequiresApi(Build.VERSION_CODES.M)
+class RecordCameraFragment : BaseTransformationFragment() {
```
IanDBird, line 56: As mentioned in the description, this is a lot of Camera2 specific code. My hope is to follow up with a PR that moves a lot of this plumbing to a `Camera2MediaSource`. That, internally, will use the new `ExternalMediaSource` but ideally reduce a lot of the complexity here that consumers would need to manage themselves.

litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCameraFragment.kt
```diff
@@ -0,0 +1,358 @@
...
+                Manifest.permission.CAMERA
+        ) == PackageManager.PERMISSION_GRANTED
+    }
+
+    /**
+     * Since we don't support cropping of the sensor or input surface, we will record to a Surface
+     * that has the same aspect ratio as the camera's sensor. We will target something similar to
+     * 720p (1280 x 720) but allow the height to increase when required for non-16:9 ratios.
+     */
+    private fun getRecordSize(sensorRect: Rect?): Size {
```
IanDBird, line 336: For now, we will require the source aspect ratio to match the target aspect ratio. For the camera specific cases, we could follow up with a couple of options:
 - Have people crop at the sensor
 - Implement some filter support for cropping

While the former is arguably simpler, it introduces the limitation that if the application is using a Preview Surface, then it will also be cropped. A crop based filter may be a better / more versatile solution.

litr/src/main/java/com/linkedin/android/litr/io/ExternalMediaSource.kt
```diff
@@ -0,0 +1,251 @@
...
+private const val DEFAULT_FRAME_RATE = 30
+private const val DEFAULT_BITRATE = 10_000_000
+private const val DEFAULT_KEY_FRAME_INTERVAL = 5
+
+/**
+ * An implementation of MediaSource, which exposes the input {@link Surface} of the encoder via a
+ * Callback. An instance of this class is expected to be both the MediaSource, and Decoder for which
+ * the pipeline is configured. This allows these components to be bypassed.
+ */
+class ExternalMediaSource(private val callback: Callback): MediaSource, Decoder {
```
IanDBird, line 35: While looking into how to implement `Camera2MediaSource`, it shows that requiring the callback via the constructor isn't that flexible. It's arguably better to change to something more like `addCallback` (or `addListener`). That's a lot more flexible.

Happy to make the change as part of this PR, or the follow up where I utilise it properly in the `Camera2MediaSource`

IanDBird:
Decided to close this PR and re-open with one that also includes `Camera2MediaSource`

PR #232, opened 2022-12-26
Implement Camera2 support as a MediaSource

IanDBird:
This PR introduces two new MediaSources:
 - `ExternalMediaSource`: This provides the owner direct access to the Surface that will be used as the input into any configured filters, followed by the encoder. The reason for this is to allow non-file based sources of video.
 - `Camera2MediaSource`: This utilises the `ExternalMediaSource` and uses Android's Camera2 APIs to render from the camera to the input Surface. This is being combined with the previously landed `AudioRecordMediaSource` that is adding the audio stream. This demonstrates a Camera to File example.

Also, one thing to note is that the ExternalMediaSource does not support cropping. It expects the source aspect ratio to match that of the file/target aspect ratio. We could either keep this as a hard constraint, or we could introduce a CroppingFilter to handle this case.

### DIFF
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCamera2Fragment.kt
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCamera2Fragment.kt
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2022 LinkedIn Corporation
+ * All Rights Reserved.
+ *
+ * Licensed under the BSD 2-Clause License (the "License").  See License in the project root for
+ * license information.
+ *
+ * Author: Ian Bird
+ */
+package com.linkedin.android.litr.demo
+
+import android.Manifest
+import android.app.Activity
+import android.app.AlertDialog
+import android.content.Context
+import android.content.pm.PackageManager
+import android.hardware.camera2.*
+import android.os.Build
+import android.os.Bundle
+import android.util.Log
+import android.view.LayoutInflater
+import android.view.SurfaceHolder
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.RequiresApi
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
+import com.linkedin.android.litr.MediaTransformer
+import com.linkedin.android.litr.demo.data.RecordCameraPresenter
+import com.linkedin.android.litr.demo.data.TargetMedia
+import com.linkedin.android.litr.demo.data.TransformationState
+import com.linkedin.android.litr.demo.databinding.FragmentCamera2RecordBinding
+import com.linkedin.android.litr.io.AudioRecordMediaSource
+import com.linkedin.android.litr.io.Camera2MediaSource
+import com.linkedin.android.litr.utils.TransformationUtil
+import java.io.File
+import java.util.*
+
+private const val TAG = "RecordCamera2Fragment"
+
+private const val REQUEST_AUDIO_AND_CAMERA_PERMISSION = 27
+
+private const val DEFAULT_CAMERA_FPS = 30
+private const val DEFAULT_TARGET_BITRATE = 5_000_000 // 5Mbps
+private const val DEFAULT_RECORD_WIDTH = 1280
+
+@RequiresApi(Build.VERSION_CODES.M)
+class RecordCamera2Fragment : BaseTransformationFragment() {
+    private lateinit var binding: FragmentCamera2RecordBinding
+
+    private lateinit var mediaTransformer: MediaTransformer
+    private var targetMedia: TargetMedia = TargetMedia()
+
+    private lateinit var cameraManager: CameraManager
+
+    private val cameraId: String? by lazy {
+        val cameraIds = cameraManager.cameraIdList
+        for (id in cameraIds) {
+            val characteristics = cameraManager.getCameraCharacteristics(id)
+            if (characteristics.get(CameraCharacteristics.LENS_FACING)
+                    == CameraCharacteristics.LENS_FACING_FRONT
+            ) {
+                continue
+            }
+
+            return@lazy id
+        }
+
+        return@lazy null
+    }
+
+    private val mediaSourceCallback = object : Camera2MediaSource.Callback {
+        override fun onDeviceReady(cameraCharacteristics: CameraCharacteristics) {
+            initPreview(cameraCharacteristics)
+
+            binding.transformationPresenter?.recordCamera(
+                    binding.audioMediaSource!!,
+                    binding.videoMediaSource!!,
+                    binding.targetMedia!!,
+                    binding.transformationState!!
+            )
+        }
+
+        override fun onFrameSkipped(frameSkipCount: Int) {
+            Log.e(TAG, "onFrameSkipped (Count: $frameSkipCount)")
+        }
+
+        override fun onError(exception: Exception) {
+            Log.e(TAG, "onError $exception")
+        }
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        cameraManager = requireContext().getSystemService(Context.CAMERA_SERVICE) as CameraManager
+        mediaTransformer = MediaTransformer(context!!.applicationContext)
+
+        // This demo fragment requires Android M or newer, in order to support reading data from
+        // AudioRecord in a non-blocking way. Let's double check that the current device supports
+        // this.
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            AlertDialog.Builder(requireContext())
+                    .setMessage(R.string.error_marshmallow_or_newer_required)
+                    .setPositiveButton(R.string.ok) { _, _ ->
+                        activity?.onBackPressed()
+                    }
+                    .show()
+            return
+        }
+
+        // Check to see what permission, if any, are required.
+        val requiredPermissions = mutableListOf<String>()
+        if (!hasAudioRecordPermission()) {
+            requiredPermissions.add(Manifest.permission.RECORD_AUDIO)
+        }
+        if (!hasCameraPermission()) {
+            requiredPermissions.add(Manifest.permission.CAMERA)
+        }
+
+        if (requiredPermissions.isNotEmpty()) {
+            ActivityCompat.requestPermissions(
+                    context as Activity,
+                    requiredPermissions.toTypedArray(),
+                    REQUEST_AUDIO_AND_CAMERA_PERMISSION
+            )
+        }
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        mediaTransformer.release()
+    }
+
+    override fun onCreateView(
+            inflater: LayoutInflater,
+            container: ViewGroup?,
+            savedInstanceState: Bundle?
+    ): View {
+        binding = FragmentCamera2RecordBinding.inflate(layoutInflater, container, false)
+
+        binding.buttonRecord.setOnClickListener { startRecording() }
+        binding.buttonStop.setOnClickListener { stopRecording() }
+
+        binding.transformationState = TransformationState()
+        binding.transformationPresenter = RecordCameraPresenter(context!!, mediaTransformer)
+        binding.audioMediaSource = AudioRecordMediaSource()
+
+        binding.videoMediaSource = Camera2MediaSource(requireContext(), cameraId!!).apply {
+            frameRate = DEFAULT_CAMERA_FPS
+            bitrate = DEFAULT_TARGET_BITRATE
+            recordWidth = DEFAULT_RECORD_WIDTH
+
+            setCallback(mediaSourceCallback)
+            addPreviewSurfaceHolder(binding.cameraView.holder)
+        }
+
+        val targetFile = File(
+                TransformationUtil.getTargetFileDirectory(requireContext().applicationContext),
+                "recorded_camera_${System.currentTimeMillis()}.mp4"
+        )
+        targetMedia.setTargetFile(targetFile)
+        binding.targetMedia = targetMedia
+
+        return binding.root
+    }
+
+    private fun startRecording() {
+        if (ActivityCompat.checkSelfPermission(
+                        requireContext(),
+                        Manifest.permission.CAMERA
+                ) == PackageManager.PERMISSION_GRANTED)
+        {
+            binding.videoMediaSource?.openCamera()
+        }
+    }
+
+    private fun initPreview(characteristics: CameraCharacteristics) {
+        val previewSize = getPreviewOutputSize(
+                binding.cameraView.display,
+                characteristics,
+                SurfaceHolder::class.java)
+        Log.i(TAG, "View finder size: ${binding.cameraView.width} x ${binding.cameraView.height}")
+        Log.i(TAG, "Selected preview size: $previewSize")
+
+        binding.cameraView.post {
+            binding.cameraView.setAspectRatio(previewSize.width, previewSize.height)
+        }
+    }
+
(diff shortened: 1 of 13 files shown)


litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
```diff
@@ -0,0 +1,253 @@
...
+        }.build()
+
+        camera.createCaptureSession(surfaces, captureStateCallback, backgroundHandler)
+    }
+
+    /**
+     * Since we don't support cropping of the sensor or input surface, we will record to a Surface
+     * that has the same aspect ratio as the camera's sensor.
+     */
+    private fun getRecordSize(sensorRect: Rect?, sensorOrientation: Int?): Size {
```
IanDBird, line 212: For now, we will require the source aspect ratio to match the target aspect ratio. For the camera specific cases, we could follow up with a couple of options:
 - Have people crop at the sensor
 - Implement some filter support for cropping

While the former is arguably simpler, it introduces the limitation that if the application is using a Preview Surface, then it will also be cropped. A crop based filter may be a better / more versatile solution.

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
```diff
@@ -61,7 +61,7 @@ class AudioRecordMediaSource(
     private var isRecording = false
 
     @Synchronized
-    fun start() {
+    fun startRecording() {
```
IanDBird, line 64: I tweaked to be consistent with the new MediaSources, and also thought being more explicit (rather than start/stop) would be better.

litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCamera2Fragment.kt
```diff
@@ -0,0 +1,213 @@
...
+private const val DEFAULT_RECORD_WIDTH = 1280
+
+@RequiresApi(Build.VERSION_CODES.M)
+class RecordCamera2Fragment : BaseTransformationFragment() {
+    private lateinit var binding: FragmentCamera2RecordBinding
+
+    private lateinit var mediaTransformer: MediaTransformer
+    private var targetMedia: TargetMedia = TargetMedia()
+
+    private val cameraManager: CameraManager by lazy {
```
IanDBird, line 54: This requires `lazy` because I am deferring it to when I need to use it, which will be *after* the context has been attached to the Fragment.

litr-demo/src/main/java/com/linkedin/android/litr/demo/view/AutoFitSurfaceView.kt
```diff
@@ -0,0 +1,71 @@
...
+                newHeight = (width / actualRatio).roundToInt()
+            }
+
+            Log.d(TAG, "Measured dimensions set: $newWidth x $newHeight")
+            setMeasuredDimension(newWidth, newHeight)
+        }
+    }
+
+    companion object {
+        private val TAG = AutoFitSurfaceView::class.java.simpleName
```
izzytwosheds, line 69: I don't think we need ` companion object` here, this can be a `const val`

litr/src/main/java/com/linkedin/android/litr/io/AudioRecordMediaSource.kt
izzytwosheds: 👍

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
```diff
@@ -0,0 +1,253 @@
...
+
+    // We support an optional SurfaceHolder that can be used as a Preview for the Camera capture.
+    private var previewSurfaceHolder: SurfaceHolder? = null
+    private val previewSurfaceHolderListener = object : SurfaceHolder.Callback {
+        override fun surfaceCreated(holder: SurfaceHolder) {
+            Log.i(TAG, "Preview Surface Created")
+            previewSurfaceHolder = holder
+        }
+
+        override fun surfaceChanged(surfaceHolder: SurfaceHolder, p1: Int, p2: Int, p3: Int) {}
```
izzytwosheds, line 56: `p1`, etc are not good parameter names, let's make them more descriptive

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
```diff
@@ -0,0 +1,253 @@
...
+    fun addPreviewSurfaceHolder(surfaceHolder: SurfaceHolder) {
+        surfaceHolder.addCallback(previewSurfaceHolderListener)
+    }
+
+    /**
+     * Opens the configured CameraDevice so that it's ready for recording. Once available, the
+     * configured Callback will be notified via onDeviceReady. Once this has done, the consumer is
+     * able to start the Transcode session.
+     */
+    @SuppressLint("MissingPermission")
```
izzytwosheds, line 143: Can we use `@RequiresPermission` (https://developer.android.com/reference/androidx/annotation/RequiresPermission) here?

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
izzytwosheds: I have used `DefaultVideoFrameRenderFilter` for cropping in the past, we could probably experiment with that here.

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
```diff
@@ -0,0 +1,253 @@
...
+
+        camera.createCaptureSession(surfaces, captureStateCallback, backgroundHandler)
+    }
+
+    /**
+     * Since we don't support cropping of the sensor or input surface, we will record to a Surface
+     * that has the same aspect ratio as the camera's sensor.
+     */
+    private fun getRecordSize(sensorRect: Rect?, sensorOrientation: Int?): Size {
+        val aspectRatio = sensorRect?.let { it.height().toFloat() / it.width().toFloat() } ?: (3f / 4f)
```
izzytwosheds, line 213: note to self: we should add an enum (or a collection of constants) for typical aspect ratios. Don't have to do this in this PR.

litr/src/main/java/com/linkedin/android/litr/io/ExternalMediaSource.kt
```diff
@@ -0,0 +1,261 @@
...
+private const val DEFAULT_FRAME_RATE = 30
+private const val DEFAULT_BITRATE = 10_000_000
+private const val DEFAULT_KEY_FRAME_INTERVAL = 5
+
+/**
+ * An implementation of MediaSource, which exposes the input {@link Surface} of the encoder via a
+ * Callback. An instance of this class is expected to be both the MediaSource, and Decoder for which
+ * the pipeline is configured. This allows these components to be bypassed.
+ */
+open class ExternalMediaSource: MediaSource, Decoder {
```
izzytwosheds, line 35: This is a question, not a change suggestion. Would it make sense to name this `CaptureMediaSource` or something similar?

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
IanDBird: I think we'd need to use both, right?
 - The `RequiresPermission` reports that the caller needs the permission
 - The `SuppressLint` avoids Android Studio trying to force that we add the permission to the litr manifest (instead requiring people who want/need to use the source to make sure they add it)

litr/src/main/java/com/linkedin/android/litr/io/ExternalMediaSource.kt
IanDBird: Yeah, I like it. I think "External" made more sense when we expected most consumers would need it directly. However, with the addition of `Camera2MediaSource`, most people may use this instead. Will rename

litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
```diff
@@ -755,7 +757,90 @@ public void stopRecording(@NonNull AudioRecordMediaSource mediaSource) {
...
+            audioTrackFormat.channelCount = 1;
+            audioTrackFormat.bitrate = 64 * 1024;
+
+            TrackTransform.Builder audioTransformBuilder = new TrackTransform.Builder(audioMediaSource,
+                    0,
+                    mediaTarget)
+                    .setTargetTrack(1)
+                    .setTargetFormat(createAudioMediaFormat(audioTrackFormat))
+                    .setEncoder(new MediaCodecEncoder())
+                    .setDecoder(new MediaCodecDecoder());
```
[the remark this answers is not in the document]
IanDBird: I can't remember exactly why I used the `MediaCodecDecoder` but I tried to change as you suggested, but it results in the following error:

```
2022-12-27 10:29:05.019 17230-17298/com.linkedin.android.litr.demo E/TransformationJob: Transformation job error
    java.nio.BufferOverflowException
        at java.nio.ByteBuffer.put(ByteBuffer.java:613)
        at java.nio.DirectByteBuffer.put(DirectByteBuffer.java:268)
        at com.linkedin.android.litr.render.PassthroughAudioProcessor.processFrame(PassthroughAudioProcessor.kt:21)
        at com.linkedin.android.litr.render.AudioRenderer.renderFrame(AudioRenderer.kt:103)
        at com.linkedin.android.litr.transcoder.AudioTrackTranscoder.queueDecodedInputFrame(AudioTrackTranscoder.java:182)
        at com.linkedin.android.litr.transcoder.AudioTrackTranscoder.processNextFrame(AudioTrackTranscoder.java:90)
        at com.linkedin.android.litr.TransformationJob.processNextFrame(TransformationJob.java:211)
        at com.linkedin.android.litr.TransformationJob.transform(TransformationJob.java:108)
        at com.linkedin.android.litr.TransformationJob.run(TransformationJob.java:77)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:463)
        at java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
        at java.lang.Thread.run(Thread.java:1012)
```

That seems to happen even if I increase the `inputBufferCapacity`. I'd suggest we keep as is, but can look into this error separately, if that's okay?

litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCamera2Fragment.kt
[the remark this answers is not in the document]
IanDBird: Sure, we switch to be consistent 👍

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
IanDBird: Actually, scrap that, you're right (and I made a typo which is why it failed locally)

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
IanDBird: Sounds good.

I thought for v1, requiring the same aspect ratio of the sensor is a requirement some people could live with. Then separately, we can iterate on it and support things like cropping.

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
IanDBird: Nice catch, thanks

litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCamera2Fragment.kt
```diff
@@ -0,0 +1,213 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        mediaTransformer = MediaTransformer(context!!.applicationContext)
+
+        // This demo fragment requires Android M or newer, in order to support reading data from
+        // AudioRecord in a non-blocking way. Let's double check that the current device supports
+        // this.
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            AlertDialog.Builder(requireContext())
+                    .setMessage("Android Marshmallow or newer required")
```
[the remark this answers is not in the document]
IanDBird: Removed hard coded strings, and also fixed the same on in the `RecordAudioFragment`

litr-demo/src/main/java/com/linkedin/android/litr/demo/RecordCamera2Fragment.kt
```diff
@@ -0,0 +1,213 @@
...
+import java.io.File
+import java.util.*
+
+private const val TAG = "RecordCamera2Fragment"
+
+private const val REQUEST_AUDIO_AND_CAMERA_PERMISSION = 27
+
+private const val DEFAULT_CAMERA_FPS = 30
+private const val DEFAULT_TARGET_BITRATE = 5_000_000 // 5Mbps
+private const val DEFAULT_RECORD_WIDTH = 1280
```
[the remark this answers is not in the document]
IanDBird: Yeah, makes sense. I'll have a think, and potentially push another update. Maybe, while we have the aspect ratio requirement, we could do something like a "maximum resolution" that we compute our target resolution within. For this example case, it would be like passing in a max resolution of "1280x1280" which is basically the max bounding box?

litr-demo/src/main/java/com/linkedin/android/litr/demo/data/TransformationPresenter.java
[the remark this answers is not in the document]
izzytwosheds: That is totally fine. It was more of a question than change suggestion.

litr/src/main/java/com/linkedin/android/litr/io/Camera2MediaSource.kt
izzytwosheds: 👍

litr/src/main/java/com/linkedin/android/litr/io/CaptureMediaSource.kt
```diff
@@ -0,0 +1,261 @@
...
+import com.linkedin.android.litr.codec.Decoder
+import com.linkedin.android.litr.codec.Frame
+import com.linkedin.android.litr.codec.PassthroughDecoder
+import com.linkedin.android.litr.exception.MediaSourceException
+import java.nio.ByteBuffer
+
+private const val TAG = "CaptureMediaSource"
+private const val DECODER_NAME = "CaptureMediaSource.Decoder"
+
+private const val MIME_TYPE_RAW = "video/raw"
```
izzytwosheds, line 25: You can add this to `MimeType` class

izzytwosheds (approved):

litr/src/main/java/com/linkedin/android/litr/io/CaptureMediaSource.kt
IanDBird: Will do. I originally resisted because it's not really a "standard" mime type, but happy to move.

IanDBird:
@izzytwosheds - That's me pushed a change to move the MimeType. If you're happy with the current PR, maybe we could land and look at any tweaks in some follow up PRs?

izzytwosheds (approved):

PR #242, opened 2023-01-19
Muxers: Add native headers required for building

IanDBird:
As discussed, we require the appropriate headers to build/link against for the bundled pre-built version of ffmpeg. While we may change in the future, this basically copies those headers into the repo. It is a lot of files, but due to the dependencies between each other, they are unfortunately mostly required.

Some potential changes in the future:
 - We require users to check out ffmpeg if wanting to use the muxer (we would have to have a different variant of the demo app to include this additional feature)
 - Instead of including the files in the repo, we could modify CMAKE to download the source directly (that matches the release tag we used) and unzip on the local device. This would require a larger download though, as **all** source would be included, not just the libraries we're building against.

Alternative approach to consider: https://github.com/linkedin/LiTr/pull/243


IanDBird:
Closing as we preferred the alternative approach: https://github.com/linkedin/LiTr/pull/243

PR #243, opened 2023-01-21
Muxers: Add native headers required for building (Alt)

IanDBird:
This is an alternative approach to https://github.com/linkedin/LiTr/pull/242.

Instead of including all the ffmpeg headers in the repo, we will download a snapshot locally. The ideal approach is that the user checks out and builds ffmpeg themselves, but this fallback allows us to ensure that a new checkout will build and use the correct headers for the bundled binaries.


litr-muxers/build.gradle
```diff
@@ -26,7 +26,7 @@ android {
     externalNativeBuild {
         cmake {
             path file('src/main/cpp/CMakeLists.txt')
-            version '3.10.2'
+            version '3.18.1'
```
IanDBird, line 29: Updated cmake required for `file(ARCHIVE_EXTRACT ...)`

izzytwosheds (approved):

PR #248, opened 2023-01-24
Muxers: Add native muxer implementation

IanDBird:
Context: https://github.com/linkedin/LiTr/pull/235

This code includes the native code for the libavformat based muxer. It provides a wrapper around the required functionality, as well as the JNI code required for exposing it to the KT/Java classes (coming in later PR).

I wasn't sure how much commenting / documentation you would like for the C++ code, so if more is required, just let me know and I can do another pass.


litr-muxers/src/main/cpp/CMakeLists.txt
```diff
@@ -48,6 +48,10 @@ else()
 
     set(ffmpeg_location "${ffmpeg_bundled_source}")
     set(ffmpeg_binaries "${CMAKE_CURRENT_SOURCE_DIR}/ffmpeg_bundled/android-libs/${ANDROID_ABI}")
+
+    # Include additional headers that would be generated via the configure (which won't be there in
+    # the downloaded artifact.
+    include_directories("${CMAKE_CURRENT_SOURCE_DIR}/ffmpeg_bundled/extra")
```
IanDBird, line 54: Hopefully the last cmake change 😅 Turned out that the `configure` auto-generated this header, so included in the repo to augment the vanilla source that we download (when the user doesn't have their own checkout).

izzytwosheds (approved):

PR #249, opened 2023-01-29
Upgrade Build Tools / SDK / Kotlin

IanDBird:
I started looking into fleshing out CameraX support but found when adding the dependency, there was a miss-match of kotlin version that reported warnings. I then went down a rabbit hole to see what would be required to upgrade. These were the changes I ended up making, so thought I would raise a PR for consideration. This also includes some fixes for the warnings coming identified by the upgraded build tools.

This change includes:
 - Updating Kotlin version from 1.4.32 -> 1.7.20
 - Updating AGP 7.1.2 -> 7.3.1
 - Updating compileSdkVersion 30 -> 33

### DIFF
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/RecordAudioPresenter.kt
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/RecordAudioPresenter.kt
@@ -9,6 +9,7 @@
  */
 package com.linkedin.android.litr.demo.data
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.Color
 import android.media.MediaMuxer
@@ -38,6 +39,7 @@ class RecordAudioPresenter(
     private val mediaTransformer: MediaTransformer
 ) : TransformationPresenter(context, mediaTransformer) {
 
+    @SuppressLint("MissingPermission")
     fun recordAudio(
         mediaSource: AudioRecordMediaSource,
         targetMedia: TargetMedia,
--- a/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/RecordCameraPresenter.kt
+++ b/litr-demo/src/main/java/com/linkedin/android/litr/demo/data/RecordCameraPresenter.kt
@@ -9,6 +9,7 @@
  */
 package com.linkedin.android.litr.demo.data
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.media.MediaMuxer
 import android.os.Build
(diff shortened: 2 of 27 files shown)


litr-demo/src/main/java/com/linkedin/android/litr/demo/data/RecordAudioPresenter.kt
```diff
@@ -38,6 +39,7 @@ class RecordAudioPresenter(
     private val mediaTransformer: MediaTransformer
 ) : TransformationPresenter(context, mediaTransformer) {
 
+    @SuppressLint("MissingPermission")
```
IanDBird, line 42: Since this is demo app code, I thought I would just suppress the warning rather than doing all the plumbing to expose the required permission. In prod code, the MediaSource correctly exposes this and the consumer can handle it accordingly.

litr-demo/src/main/java/com/linkedin/android/litr/demo/data/RecordCameraPresenter.kt
```diff
@@ -31,6 +32,7 @@ class RecordCameraPresenter(
     private val mediaTransformer: MediaTransformer
 ) : TransformationPresenter(context, mediaTransformer) {
 
+    @SuppressLint("MissingPermission")
```
IanDBird, line 35: Since this is demo app code, I thought I would just suppress the warning rather than doing all the plumbing to expose the required permission. In prod code, the MediaSource correctly exposes this and the consumer can handle it accordingly.

litr-demo/src/main/java/com/linkedin/android/litr/demo/fragment/CameraSizes.kt
```diff
@@ -32,32 +30,18 @@ class SmartSize(width: Int, height: Int) {
 @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
 val SIZE_1080P: SmartSize = SmartSize(1920, 1080)
 
-/** Returns a [SmartSize] object for the given [Display] */
-@RequiresApi(Build.VERSION_CODES.LOLLIPOP)
-fun getDisplaySmartSize(display: Display): SmartSize {
-    val outPoint = Point()
-    display.getRealSize(outPoint)
```
IanDBird, line 39: This is deprecated. Since it was only there to handle devices with "smaller than 1080p screens" and is only demo app code, I decided to simplify the code and just remove it.

izzytwosheds (approved):
Thank you!

PR #263, opened 2023-08-02
Fix potential buffer overflow issue

Nailik:
I found this issue while implementing a small [resampler](https://github.com/Nailik/AndroidResampler) based on this project.

**Description**
`AudioRenderer`and `AudioOverlayFilter`use:
```
val estimatedTargetSampleCount = ceil(sourceSampleCount * samplingRatio).toInt()
 val targetBufferCapacity = estimatedTargetSampleCount * channelCount * BYTES_PER_SAMPLE
```
to calculate the size of the target buffer.
In case this estimation is too small (happens on big differences in samplerate) this results in 2 issues.

**Issue**
`processAudioFrame` in `audio-processor.cpp` doesn't know about the size, resulting in memory access vialoation.

**Fix**
Drop audio when target buffer capacity is reached.

**Issue**
`targetFrame.buffer.limit(targetBufferSize)` throws an `IllegalArgumentException`.

**Fix**
Take min `min(targetBufferSize, targetFrame.buffer.capacity())`

### DIFF
--- a/litr/src/main/cpp/audio-processor.cpp
+++ b/litr/src/main/cpp/audio-processor.cpp
@@ -52,7 +52,8 @@ Java_com_linkedin_android_litr_render_OboeAudioProcessor_processAudioFrame(
         jobject,
         jobject jsourceBuffer,
         jint sampleCount,
-        jobject jtargetBuffer) {
+        jobject jtargetBuffer,
+        jint jtargetBufferSize) {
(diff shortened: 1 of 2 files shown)


izzytwosheds (approved):
Nice fix. Thank you!

Nailik:
@izzytwosheds i fixed a small issue you could review again.

izzytwosheds (approved):

PR #270, opened 2024-01-12
Fixes a crash in demo app due to a null pointer

vamshi-dhulipala:
Update the logic in fragment_transcode_video_gl.xml to use a default value of false for the data binding variable 'enableNativeMuxer' whenever it is null.


nlgtuankiet:
@izzytwosheds 
I fate this issue myself, could you please consider merge this?

izzytwosheds (approved):

Issue #269, opened 2023-12-05
TranscoderUtils.java line 134


Ma7moudHatem:
I have the following crash on my app caused by LiTr lib:

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'long java.lang.Long.longValue()' on a null object reference
       at android.media.MediaFormat.getLong(MediaFormat.java:855)
       at com.linkedin.android.litr.utils.TranscoderUtils.estimateVideoTrackBitrate(TranscoderUtils.java:134)
       ...
```

Could you please fix it as it affected many of my app users?

**Extra Info:**
LiTr version: com.linkedin.android.litr:litr:1.5.5
Android OS: Different OS versions
Android device: Different Android devices

PR #271, opened 2024-01-18
Fixes a null pointer crash (Github Issue #269)

vamshi-dhulipala:
1. Fixes a null pointer in TranscoderUtils when retrieving the duration value from a track by checking if the field is present before accessing it.
2. Updated logic of estimateVideoTrackBitrate API in TranscoderUtils to utilize the source media file duration in cases where track specific duration value is null.
3. Updated logic of createTargetMediaFormat API in MediaTransformer class to use a default bit rate of 10_000_000 when we fail to estimate a valid bitrate from the source media track.
4. Added unit tests

### DIFF
--- a/litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
+++ b/litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
@@ -174,14 +174,14 @@ public final class TranscoderUtils {
      * shall be returned.
      *
      * @param mediaSource {@link MediaSource} which contains the video track
-     * @param videoTrack Track whose duration needs to be determined
+     * @param videoTrackFormat MediaFormat of the track whose duration needs to be determined
      *
      * @return Duration of the given video track in seconds.
      */
     @VisibleForTesting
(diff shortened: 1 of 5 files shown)


litr/src/main/java/com/linkedin/android/litr/utils/TranscoderUtils.java
```diff
@@ -165,6 +168,23 @@ public static int estimateVideoTrackBitrate(@NonNull MediaSource mediaSource, in
...
+     * duration value in its meta data, then the duration value associated with the media source
+     * shall be returned.
+     *
+     * @param mediaSource {@link MediaSource} which contains the video track
+     * @param videoTrack Track whose duration needs to be determined
+     *
+     * @return Duration of the given video track in seconds.
+     */
+    @VisibleForTesting
+    static float estimateVideoTrackDuration(MediaSource mediaSource, MediaFormat videoTrack) {
```
izzytwosheds, line 182: nip: `videoTrackFormat` for the second parameter

izzytwosheds (approved):

